PR #1, opened 2016-07-20
Update C# tests project to use wildcards

LogicAndTrick:
I've updated the C# test project a bit, but I'm not sure if it'll break the CI (and I don't have a Linux box that I can test on at the moment). Any pointers on the best way to approach it? Should I just commit and see if it works?

(the project does build on mono for Windows, so I think it should be okay)


spec/csharp/.gitignore
```diff
@@ -0,0 +1,4 @@
+packages/
+.vs/
+bin/
+obj/
```
GreyCat: 1. It is generally a bad idea to omit newlines at the end of file. I'm not sure about C# style guide, but it's almost universally frowned upon in C/C++, Java, most scripting languages and in git itself.

2. Shouldn't we just put https://github.com/github/gitignore/blob/master/VisualStudio.gitignore there?

spec/csharp/kaitai_struct_csharp_tests/CommonSpec.cs
```diff
@@ -1,9 +1,14 @@
...
 namespace Kaitai
 {
     public class CommonSpec
     {
-        public String SRC_DIR = "src/";
+        public String SourceFile(string filename)
+        {
+            var currentDir = Path.GetDirectoryName(typeof(CommonSpec).Assembly.Location);
+            return Path.Combine(currentDir, "src", filename);
+        }
```
GreyCat: I don't quite get what it does, but if it tries to derive location of binary input files (i.e. `src`) from the location of assembly somehow, probably it's a bad idea. That `src` is usually located in `tests/src`, while assembly ends up in `tests/spec/csharp/kaitai_struct_csharp_tests/bin/Debug/kaitai_struct_csharp_tests.dll`.

In ideal world, we probably shall supply the location of test input files using some external mechanism, like environment variable, interpreter property, command line arguments or something like that. Test running automation scripts (like `tests/run-csharp`) actually have the config (`tests/config`) which should be the only place which actually stores path to `src`, and they should supply this path to the tests as a piece of external config.

`src` was a lousy naming choice, by the way. Naming test inputs as "sources" in a system that deals with generation of source files from other source files, was, well, a bad idea of mine :) Hope to fix it someday.

spec/csharp/kaitai_struct_csharp_tests/kaitai_struct_csharp_tests.csproj
```diff
@@ -35,22 +35,40 @@
...
     </Compile>
+    <Compile Include="tests\*.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="..\..\..\src\*.bin">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      <Link>src\%(Filename)%(Extension)</Link>
+    </None>
+    <None Include="packages.config" />
+  </ItemGroup>
```
GreyCat: Ah, here's why you refer to `src`, starting from an assembly...

GreyCat:
> I've updated the C# test project a bit, but I'm not sure if it'll break the CI (and I don't have a Linux box that I can test on at the moment). Any pointers on the best way to approach it? Should I just commit and see if it works?

I'd say go ahead and try it. You can try running `compile-all` and `run-csharp` on Windows box too, probably it will be fine with [msys](http://www.mingw.org/wiki/msys) or some other UNIX shell tools implementation. Or even without them - hopefully successful build VS solution and a run of NUnit runners would guarantee that everything would be ok. And even it's not, we're talking about unstable, WIP stuff here, things are known to break occasionally. Even complete crash of CI is ok :) In the worst case, we can always just revert to previous revisions.

LogicAndTrick:
Alright, I'll give it a shot. Just so I'm sure about the process to trigger CI, is this process right?

1. Merge this PR into master
2. Pull the submodule in the parent `kaitai_struct` project
3. Commit + push parent project
4. CI automatically builds

GreyCat:
That is correct. If you've used "recommended" checkout scheme, i.e. `git clone --recursive` for parent project, you don't need to actually pull anything for step 2, you just need to add + commit + push information about new changes.

LogicAndTrick:
Looks like I don't have push access to the `kaitai_struct` project. Can you pull the changes in please?

GreyCat:
Fixed the rights. Feel free to pull it yourself as much as you need :)

I was kind of misguided on how github's permissions work - in fact, being a "team member" does not grant you any access (beyond reading, which you have anyway for public projects) automatically, one has either to belong to a "team" with write access, or be a "collaborator" on a project.

LogicAndTrick:
Thanks for that. Looks like the build works as expected but the few test cases with name conflicts are breaking the build. I'll use the hard-coded project file for now, and revert back to this once those tests have been updated (or partial compiling for C# is working).

GreyCat:
Yeah, it [looks like](https://travis-ci.org/kaitai-io/kaitai_struct/builds/146079396#L2407) we can't use wildcards, at least with `compiled/` stuff so far due to these syntax errors :(

LogicAndTrick:
Okay, wildcards aren't going to work the way I hoped they would. Reverting back to the fully hard-coded project...

LogicAndTrick:
It works! I've ported the compileable tests and now C# has a 75% pass rate. http://kaitai.io/ci/

GreyCat:
That's super cool :) I'll make a note on Twitter for that :)

PR #2, opened 2016-07-24
Change test cases to avoid name conflicts in C#

LogicAndTrick:
These 5 unit tests have a naming conflict in C#, the fix mentioned in kaitai-io/kaitai_struct_compiler#2 was to rename the tests for now. I've appended `_obj` onto the conflicting class names to fix the issue. I don't think any unit tests need to change because they shouldn't refer to the class by name.

### DIFF
--- a/formats/instance_io_user.ksy
+++ b/formats/instance_io_user.ksy
@@ -9,7 +9,7 @@ seq:
     repeat: expr
     repeat-expr: qty_entries
   - id: strings
-    type: strings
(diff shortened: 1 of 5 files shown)


GreyCat:
Looks like good idea :) Let's try it!

LogicAndTrick:
Thanks. I've just pushed the rest of the C# unit tests, so hopefully they'll all work now.

PR #3, opened 2016-07-26
Floating point KSY test case (C#, Java, Ruby)

LogicAndTrick:
This is a basic test case for FP support. I've ported it to C#, Java, and Ruby. (haven't managed to get Python or JS environments going yet)

To run these, you'll need to pull the latest updates from the runtimes and the compiler.


spec/ruby/floating_points_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+RSpec.describe FloatingPoints do
+  it 'parses test properly' do
+    r = FloatingPoints.from_file('src/floating_points.bin')
+    
+    delta = 1e-6
+
+    expect(r.single_value).to be_within(delta).of 1.2345
+    expect(r.single_value_be).to be_within(delta).of 1.2345
+    expect(r.double_value).to be_within(delta).of 123.456
+    expect(r.double_value_be).to be_within(delta).of 123.456
```
GreyCat: These probably has to be exact matches. It's not like there are any machine-dependent or floating point precision-dependent in just reading the value in its serialized form.

GreyCat:
Thanks! Don't worry about Python & JS - I'll do the tests and the runtimes.

spec/ruby/floating_points_spec.rb
LogicAndTrick: I probably should have used numbers that can be exactly represented as a FP value, my bad. Ruby was being a bit strange on these ones:

     Failure/Error: expect(r.single_value).to eq 1.2345

       expected: 1.2345
            got: 1.2345000505447388

I've not done much Ruby programming so I'm not sure if there's some way to change the behaviour without using a delta.

GreyCat: Actually, it doesn't have much with exactly Ruby. It's how numbers are stored internally, there is actually no way to store exact value of stuff like 1.2345 in binary numeric system, so everyone will resort to using an approximation:

## Ruby

```ruby
def pack_and_unpack(x); [x].pack('f').unpack('f')[0]; end

pack_and_unpack(1) # => 1.0
pack_and_unpack(1.25) # => 1.25
pack_and_unpack(1.26) # => 1.2599999904632568
pack_and_unpack(1.234) # => 1.2339999675750732
pack_and_unpack(1.2345) # => 1.2345000505447388
```

## Python

```python
def pack_and_unpack(x): return unpack('f', pack('f', x))[0]

pack_and_unpack(1) # => 1.0
pack_and_unpack(1.25) # => 1.25
pack_and_unpack(1.26) # => 1.2599999904632568
pack_and_unpack(1.234) # => 1.2339999675750732
pack_and_unpack(1.2345) # => 1.2345000505447388
```

For both languages, 1.2345000505447388 != 1.2345. Actually, there is no such thing as precise "1.2345" anyway - it's just language's fancy rendition of infinite periodic non-integer value that represents it:

1.0011(1100000010000011000100100110111010010111100011010100111111011111001110110110010001011010000111001010)...

Both Ruby & Python internal types use double for floating point, thus dumping an into single precision and then converting them (internally) to double zeroes out some valuable mantissa bits, and thus this "fancy rendition" magic fails to work. I believe that would be the same with C, C#, Java, etc, unless you dump & retrieve it in exactly the same precisions.

You might want to choose floats that have exact, non-periodic rendition in binary for these tests. And maybe throw in one with the infinite period in, but clearly marked as so ;)

LogicAndTrick: Ah, I forgot about the internal representation always being double for languages like JS and Ruby. I didn't run into the problem on the C# tests because I was using `1.2345f` to make sure the representation is the same. It's probably not needed for Java or C++ either, they have native support for single precision floats.

I'll tweak the test later to use values that can be represented exactly in a float/double, then a delta won't be needed for any language.

LogicAndTrick:
I've just changed this test to use values that can be exactly represented by floats. See 3bb666e0817e29044b88004baec07672b1870369.

GreyCat:
Yeah :) I've done JS tests, so it looks like we've almost packed for floating points - only Java compiling thing remains.

LogicAndTrick:
Do you have any ideas with the Java thing? Would something like this work in the `JavaTranslator`?

```
  override def numericBinOp(left: Ast.expr, op: Ast.operator, right: Ast.expr) =
    (detectType(left), detectType(right)) match {
      case (_: FloatType, _: IntType) | (_:IntType, _: FloatType) =>
        s"(((double) ${translate(left)}) ${binOp(op)} ((double) ${translate(right)}))"
      case _ =>
        super.numericBinOp(left, op, right)
    }
```

GreyCat:
We'd probably need more fine-grained translation. For example, if Java disallows us to do binary operation between `float a` and `double b`, then probably the correct way do it would be `((double) a) + b`.

GreyCat:
Checked this one. Actually, the problem is slightly different:

```java
    private Double singleValuePlusInt;
    // ...
    singleValuePlusInt = (singleValue() + 1);
```

Java is perfectly ok with doing `float` + `int` or `int` + `float` (it generates `float`), but this particular clause doesn't work because autoboxing doesn't work this way. One can assign `float` into `double` (it will be typecasted), one can assign `double` into `Double` (it would be autoboxed), but one can't just do both at once and assign `float` into `Double`.

The easiest workaround I can think of might look like that:

```java
    private Double singleValuePlusInt;
    // ...
    double _tmp = (singleValue() + 1);
    singleValuePlusInt = _tmp;
```

LogicAndTrick:
Thanks for checking it out, I'll give that a shot.

GreyCat:
Oh, I'm already on it :) Let me finish :)

LogicAndTrick:
Ok, cool! It's about time for me to get some sleep anyway :)

GreyCat:
Thanks for all of your input :) Have a good night then :)

PR #4, opened 2016-07-27
Partial compilation script for C#

LogicAndTrick:
Similar to the `run-java` script, but for C#. Should resolve kaitai-io/kaitai_struct#6. Tested in Bash on Windows - I don't do a lot of Bash scripting so please check it's okay!

There's a few things to note:

- `msbuild`/`xbuild` should be used the build the project/solution, because it can contain commands aside from just the compile step (such as copying files)
- `xbuild` doesn't like when a file is missing, so I'm just clearing the contents of the file instead
- Invalid files are found by running `mcs` and scanning stderr for the files that contain errors
- Once .NET Core has been changed to work with `.csproj` files ([Q1 2017](https://blogs.msdn.microsoft.com/dotnet/2016/07/15/net-core-roadmap/)) it might be a good idea to switch from Mono to .NET Core for the build

The output of `mcs` has the file name for each error, each line is chopped at the first `(` character and that file is erased if it's not already blank.

    /kaitai_struct/tests/compiled/csharp/DefaultBigEndian.cs(32,246): error CS1525: Unexpected symbol `end-of-file'
    echo $line | cut -d '(' -f 1
    /kaitai_struct/tests/compiled/csharp/DefaultBigEndian.cs

### DIFF
--- a/run-csharp
+++ b/run-csharp
@@ -5,72 +5,53 @@
 CSHARP_TEST_FORMATS_DIR=$FORMATS_COMPILED_DIR/csharp
 CSHARP_SPEC_DIR=spec/csharp
 CSHARP_TEST_OUT_DIR="$TEST_OUT_DIR"/csharp
+CSHARP_COMPILE_LOG="$CSHARP_TEST_OUT_DIR/_compile.log"
 
 rm -rf "$CSHARP_TEST_OUT_DIR"
 mkdir -p "$CSHARP_TEST_OUT_DIR"
 
 # Try and build the project first - if there's no compile errors, we can continue straight away
 echo "run-csharp: building project"
-xbuild "$CSHARP_SPEC_DIR/kaitai_struct_csharp_tests.sln" >> "$CSHARP_TEST_OUT_DIR/_compile.log" || {
+xbuild "$CSHARP_SPEC_DIR/kaitai_struct_csharp_tests.sln" >> "$CSHARP_COMPILE_LOG" || {
 
     # Some formats failed to build - find the failing files and erase them
-	echo "run-csharp: some formats failed to compile, checking each file individually"
+    echo "run-csharp: some formats failed to compile, checking each file individually"
     
     # Find the invalid files and wipe them - we do this by scanning the error list from the compiler
     mcs \
         -t:library \
         -fullpaths \
-        -o:$CSHARP_TEST_OUT_DIR/_cs_compile_temp.dll \
+        -warn:0 \
         -r:"$CSHARP_SPEC_DIR/packages/NUnit.3.4.1/lib/net45/nunit.framework.dll" \
         "$CSHARP_RUNTIME_DIR/*.cs" \
         "$CSHARP_SPEC_DIR/kaitai_struct_csharp_tests/CommonSpec.cs" \
         "$CSHARP_TEST_FORMATS_DIR"/*.cs \
-            2>&1 >> "$CSHARP_TEST_OUT_DIR/_compile.log" | \
-            while read -r line ; do
-                # e.g. path/to/file.cs(1): Unexpected symbol
-                # Grab everything before the first '(' character
-                FILENAME=$(echo $line | cut -d '(' -f 1)
-                if [ -s $FILENAME ]
-                then
-                    NAME=$(basename "$FILENAME")
-                    echo "$NAME failed to compile, removing"
-                    > $FILENAME
-                fi
-            done
+            2>&1 >> "$CSHARP_COMPILE_LOG" | \
+            fgrep '(' | cut -d'(' -f1 | xargs -d '\n' truncate -c -s0
     
-	echo "run-csharp: removing tests that are no longer valid"
+    echo "run-csharp: removing tests that are no longer valid"
     
     # Now that the invalid spec classes are gone, we need to remove the matching test cases
     mcs \
         -t:library \
         -fullpaths \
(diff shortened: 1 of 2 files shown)


run-csharp
```diff
@@ -2,10 +2,71 @@
...
+        -r:"$CSHARP_SPEC_DIR/packages/NUnit.3.4.1/lib/net45/nunit.framework.dll" \
+        "$CSHARP_RUNTIME_DIR/*.cs" \
+        "$CSHARP_SPEC_DIR/kaitai_struct_csharp_tests/CommonSpec.cs" \
+        "$CSHARP_TEST_FORMATS_DIR"/*.cs \
+            2>&1 >> "$CSHARP_TEST_OUT_DIR/_compile.log" | \
+            while read -r line ; do
+                # e.g. path/to/file.cs(1): Unexpected symbol
+                # Grab everything before the first '(' character
+                FILENAME=$(echo $line | cut -d '(' -f 1)
+                if [ -s $FILENAME ]
```
GreyCat: Definitely will abort with an error when there's no filename - probably it should be `-s "$FILENAME"`.

run-csharp
```diff
@@ -2,10 +2,71 @@
...
+                then
+                    NAME=$(basename "$FILENAME")
+                    echo "$NAME failed to compile, removing"
+                    > $FILENAME
+                fi
+            done
```
GreyCat: I don't really see a point in running `cut` for every line. Something along the lines of

```shell
... | fgrep '(' | cut -d'(' -f1 | xargs -n1 basename
```

will probably get you a list of failed sources.

run-csharp
```diff
@@ -2,10 +2,71 @@
...
+                # e.g. path/to/file.cs(1): Unexpected symbol
+                # Grab everything before the first '(' character
+                FILENAME=$(echo $line | cut -d '(' -f 1)
+                if [ -s $FILENAME ]
+                then
+                    NAME=$(basename "$FILENAME")
+                    echo "The $NAME test was removed as there was no matching spec"
+                    > $FILENAME
+                fi
+            done
```
GreyCat: This stuff is used twice - maybe it's worth to pack it in a function like `process_log_and_empty_files_that_failed_to_build` and reuse it?

run-csharp
```diff
@@ -2,10 +2,71 @@
...
+CSHARP_TEST_OUT_DIR="$TEST_OUT_DIR"/csharp
 
-# Compile everything
-xbuild "$CSHARP_SPEC_DIR/kaitai_struct_csharp_tests.sln"
+rm -rf "$CSHARP_TEST_OUT_DIR"
+mkdir -p "$CSHARP_TEST_OUT_DIR"
+
+# Try and build the project first - if there's no compile errors, we can continue straight away
+echo "run-csharp: building project"
+xbuild "$CSHARP_SPEC_DIR/kaitai_struct_csharp_tests.sln" >> "$CSHARP_TEST_OUT_DIR/_compile.log" || {
```
GreyCat: _compile.log is always appended to here, never cleared. Probably it's a good idea to do `rm -f "$CSHARP_TEST_OUT_DIR/_compile.log"` somewhere at the beginning? Also, if `"$CSHARP_TEST_OUT_DIR/_compile.log"` is used more than once, probably it's worth to make a new variable for it?

LogicAndTrick: The whole `$CSHARP_TEST_OUT_DIR` directly is deleted on line 9, that should handle the append/clear problem. You're right about a variable for the log file though, I'll do that.

run-csharp
LogicAndTrick: Definitely pushing what I know about shell scripting here! Being a Windows user I'm not too familiar with these commands, I'll look those up and see what I can manage.

Will I still need to `while read -r line` after doing that, to erase the files? Or is there some way I can do that without a loop as well?

Like this (just guessing for now):
```
[compile command] | fgrep '(' | cut -d'(' -f1 | while read -r line ; do
    if [ -s "$line" ]
    then
        NAME=$(basename "$line")
        echo "$NAME failed to compile, removing"
        > $line
    fi
done
```

GreyCat: You don't actually remove the files, you replace them with empty ones, do you? I'd rather kept it a bit simpler and more readable:

```shell
COMPILE_LOG1="$CSHARP_TEST_OUT_DIR/_compile1.log"

mcs blah blah 2>&1 >"$COMPILE_LOG1" || {
    fgrep '(' "$COMPILE_LOG1" | cut -d'(' -f1 >"$FAILED1"

    echo "Failed to compile, removing broken files:"
    cat "$FAILED1"

    cat "$FAILED1" | xargs truncate -s0
}
```

Even so, I don't really see much point in listing emptied files other than debugging the script - for informative purposes, they all got listed above anyway. This might make it as short as:
```shell
COMPILE_LOG1="$CSHARP_TEST_OUT_DIR/_compile1.log"

mcs blah blah 2>&1 >"$COMPILE_LOG1" ||
    fgrep '(' "$COMPILE_LOG1" | cut -d'(' -f1 | xargs truncate -s0
```

LogicAndTrick: Thanks for the tips, that's a lot neater. I'll test it out.

LogicAndTrick:
I've made those changes, hopefully it's good to go now. I'm passing `-d '\n'` to `xargs` which tells it to ignore backslashes (otherwise it doesn't work for Windows\file\paths).

arekbulski:
Can I delete the PR branch?

GreyCat:
Yeah, thanks, I've actually just deleted it.

PR #5, opened 2016-08-15
Fixed HelloWorld PHP test

markbook2:


GreyCat:
I've merged this one, but it ends up with

```
EPHP Fatal error:  Cannot declare class HelloWorld, because the name is already in use in /home/greycat/git/kaitai_struct/tests/compiled/php/HelloWorld.php on line 19
```

at least on my box. Contents of `compiled/php/HelloWorld.php` are as follows:

```php
<?php
// This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild

namespace \Kaitai\Struct\Tests;

class HelloWorld extends \Kaitai\Struct\Struct {

    public function __construct(\Kaitai\Struct\Stream $io, \Kaitai\Struct\Struct $parent = null, HelloWorld $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }
    private function _parse() {
        $this->one = $this->_io->readU1();
    }
    protected $one;
    protected $_root;
    protected $_parent;
    public function one() { return $this->one; }
}
```

Any ideas what might be the problem?

markbook2:
I have left the comment [here](https://github.com/kaitai-io/kaitai_struct/issues/16#issuecomment-239759375)

PR #6, opened 2016-08-15
Converted Java tests to PHP and fixed DefaultBigEndianTest

markbook2:


markbook2:
I would like to ensure that the commit c61ac02 should be here, may be I need create new pull request with this commit removed as I did `git rebase`. It is in the list here: https://github.com/morpho-os/kaitai_struct_tests/commits/master, it is older, but it present after the more fresh commit a2f787748c79849a591a6b342591a88e8292182a.

markbook2:
So valid sequence of commits is:
1) run-all: added run-php
2) Converted Java tests to PHP
3) Fixed the DefaultBigEndianTest

but if you merge this pull request I think it will be:
1) run-all: added run-php
2) Converted Java tests to PHP
3) run-all: added run-php
4) Fixed the DefaultBigEndianTest
for some reason, probably due `git rebase`, the `run-all: added run-php` is present twice.

GreyCat:
That's ok, it applies cleanly and does not create duplicate lines or anything like that.

PR #7, opened 2016-08-24
Fixed the RepeatUntilS4Test

markbook2:
The file RepeatUntilS4.php has the following line:
```php
        $this->afterall = $this->_io->readStrz("ASCII", 0, false, true, true);
```
it should be changed to the:
```php
        $this->afterall = $this->_io->readStrz("ASCII", "\x00", false, true, true);
```
The 0 is integer, but actual byte's value will be a character with the ASCII code 0, not an integer. The character with the code 0 can be encoded as string "\x00" in PHP.

Of course I could add a check and conversion to the character from the ASCII code:
```php
if (is_int($terminator)) {
    $terminator = chr($terminator);
}
```
But would it make a sense? I don't think that this is required as the bytes represented as hex strings in PHP.

### DIFF
--- a/spec/php/RepeatUntilS4Test.php
+++ b/spec/php/RepeatUntilS4Test.php
@@ -3,7 +3,7 @@ namespace Kaitai\Struct\Tests;
 
 class RepeatUntilS4Test extends TestCase {
     public function testRepeatUntilS4() {
-        $r = RepeatUntilS4::fromFile(self::SRC_DIR_PATH . "/repeat_n_struct.bin");
+        $r = RepeatUntilS4::fromFile(self::SRC_DIR_PATH . "/repeat_until_s4.bin");
 
         $this->assertEquals([0x42, 0x1337, -251658241, -1], $r->entries);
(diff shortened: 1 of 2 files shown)


markbook2:
The php runtime has been updated also.

GreyCat:
As far as I know, there is no "character" type in PHP, only strings - and we don't have "bytes" here, only one byte. Using string for just one byte is probably an overkill. In most other languages, we just use a byte-sized integer to represent that single character.

Generating literals for readStrz call is a bad idea: actually, we don't have an immediate value to begin with. Consider this example:

```yaml
seq:
  - id: term_byte
    type: u1
  - id: some_str
    type: strz
    terminator: term_byte
```

Thus, it would generate a call like

```php
$this->someStr = $this->_io->readStrz("ASCII", $this->termByte, false, true, true);
```

Of course, I can add `chr()` right into the call:

```php
$this->someStr = $this->_io->readStrz("ASCII", chr($this->termByte), false, true, true);
```

but it's probably easier to add it right into the `readStrz` implementation, i.e. without any `if`s?

PR #8, opened 2016-08-25
Repeat-until complex test

LogicAndTrick:
A more complex test case for the repeat-until feature, using complex types and multiple loops in the same type. I haven't gotten a JS environment going yet, so the JS test might be incorrect.

This test fails for Java and C# (also probably C++) because the loop variable is declared multiple times in the same scope:

```cs
TypeU1 M_;
do { ... } while ( ... );
TypeU2 M_;
do { ... } while ( ... );
byte M_;
do { ... } while ( ... );
```

Quick and dirty fix might be to just push the scope an extra level before looping (though not all languages are this convenient):

```cs
{
    TypeU1 M_;
    do { ... } while ( ... );
}
{
    TypeU2 M_;
    do { ... } while ( ... );
}
{
    byte M_;
    do { ... } while ( ... );
}
```

Cleaner fix might be to use a custom variable name for each loop:

```cs
TypeU1 M_it_first;
do { ... } while ( ... );
TypeU2 M_it_second;
do { ... } while ( ... );
byte M_it_third;
do { ... } while ( ... );
```


GreyCat:
Thanks! I'm still hesitating if it's worth to generate unique iterator variable names. While it's technically possible, it seems that `{}` scoping works with C++, Java and C# — thus it (1) solves the problem, (2) actually, loops' iterators are *meant* to have a loop-specific scope. In the vast majority of languages, stuff like

```c
for (int i = 0; i < len; i++) { ... }
```

means that `i` is visible only in the loop and it's completely ok to do stuff like:

```c
for (int i = 0; i < len1; i++) { ... }
for (int i = 0; i < len2; i++) { ... }
for (int i = 0; i < len3; i++) { ... }
```

LogicAndTrick:
Good point, that makes a lot of sense. I can't think of any strongly typed languages that wouldn't let you do some kind of manual scope manipulation, so the unique iterator variable names shouldn't be needed.

GreyCat:
Looks like I've fixed it for Java, C# and C++.

PR #13, opened 2016-09-01
Fixed PositionInSeq, InstanceIoUser, InstanceStdArray, IfStruct

markbook2:
PositionInSeq:
```php
<?php
// This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild

namespace Kaitai\Struct\Tests;

class PositionInSeq extends \Kaitai\Struct\Struct {

    public function __construct(\Kaitai\Struct\Stream $io, \Kaitai\Struct\Struct $parent = null, PositionInSeq $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }

    private function _parse() {
        $this->numbers = [];
        $n = $this->header()->qtyNumbers();
        for ($i = 0; $i < $n; $i++) {
            //array_push($this->numbers, $this->_io->readU1());
            $this->numbers[] = $this->_io->readU1();
        }
        if (count($this->numbers) !== $n) {
            // The array_slice() can be used also, see comments in the InstanceIoUser, for the similar chunk of the code.
            throw new \OutOfBoundsException();
        }
    }

    protected $header;

    public function header(): PositionInSeq\HeaderObj {
        if ($this->header !== null)
            return $this->header;
        $_pos = $this->_io->pos();
        $this->_io->seek(16);
        $this->header = new PositionInSeq\HeaderObj($this->_io, $this, $this->_root);
        $this->_io->seek($_pos);
        return $this->header;
    }

    protected $numbers;

    public function numbers() { return $this->numbers; }
}

namespace Kaitai\Struct\Tests\PositionInSeq;

class HeaderObj extends \Kaitai\Struct\Struct {
    public function __construct(\Kaitai\Struct\Stream $io, \Kaitai\Struct\Tests\PositionInSeq $parent = null, \Kaitai\Struct\Tests\PositionInSeq $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }
    private function _parse() {
        $this->qtyNumbers = $this->_io->readU4le();
    }
    protected $qtyNumbers;
    public function qtyNumbers() { return $this->qtyNumbers; }
}
```
InstanceIoUser:
```php
<?php
// This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild

namespace Kaitai\Struct\Tests;

class InstanceIoUser extends \Kaitai\Struct\Struct {
    public function __construct(\Kaitai\Struct\Stream $io, \Kaitai\Struct\Struct $parent = null, InstanceIoUser $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }

    private function _parse() {
        $this->qtyEntries = $this->_io->readU4le();
        $this->entries = [];//new ArrayList <InstanceIoUser\Entry>((int)($this->qtyEntries));
        $n = $this->qtyEntries();
        for ($i = 0; $i < $n; $i++) {
            // $this->entries[] does the same thing as array_push() does, but without function call.
            $this->entries[] = new InstanceIoUser\Entry($this->_io, $this, $this->_root);
        }
        // PHP does not have a special construct for initialization of the size of an array.
        // So we use an `if` with throwing of an exception.
        // Instead of throwing the \OutOfBoundException we could use another approach:
        // apply the array_slice() and therefore trim an array to the required length,
        // e.g.: $this->entries = array_slice($this->entries, 0, $this->qtyEntries);
        // but in this case we will not know that some elements have been trimmed.
        if (count($this->entries) !== $n) {
            throw new \OutOfBoundsException();
        }

        $this->_raw_strings = $this->_io->readBytesFull();
        $io = new \Kaitai\Struct\Stream($this->_raw_strings);
        $this->strings = new InstanceIoUser\StringsObj($io, $this, $this->_root);
    }
    protected $qtyEntries;
    protected $entries;
    protected $strings;
    protected $_raw_strings;
    public function qtyEntries() { return $this->qtyEntries; }
    public function entries() { return $this->entries; }
    public function strings() { return $this->strings; }
    public function _raw_strings() { return $this->_raw_strings; }
}

namespace Kaitai\Struct\Tests\InstanceIoUser;

class Entry extends \Kaitai\Struct\Struct {

    public function __construct(\Kaitai\Struct\Stream $io, \Kaitai\Struct\Tests\InstanceIoUser $parent = null, \Kaitai\Struct\Tests\InstanceIoUser $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }
    private function _parse() {
        $this->nameOfs = $this->_io->readU4le();
        $this->value = $this->_io->readU4le();
    }
    protected $name;
    public function name(): string {
        if ($this->name !== null)
            return $this->name;

        $io = $this->_root->strings()->_io;
        $_pos = $io->pos();
        $io->seek($this->nameOfs);
        $this->name = $io->readStrz("UTF-8", 0, false, true, true);
        $io->seek($_pos);
        return $this->name;
    }
    protected $nameOfs;
    protected $value;
    public function nameOfs() { return $this->nameOfs; }
    public function value() { return $this->value; }
}
class StringsObj extends \Kaitai\Struct\Struct {
    public function __construct(\Kaitai\Struct\Stream $io, \Kaitai\Struct\Tests\InstanceIoUser $parent = null, \Kaitai\Struct\Tests\InstanceIoUser $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }
    private function _parse() {
        $this->str = [];
        while (!$this->_io->isEof()) {
            // array_push() can be used also, but the $this->str[] should be faster
            $this->str[] = $this->_io->readStrz("UTF-8", 0, false, true, true);
        }
    }
    protected $str;
    public function str() { return $this->str; }
}
```
InstanceStdArray:
```php
<?php
// This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild

namespace Kaitai\Struct\Tests;

class InstanceStdArray extends \Kaitai\Struct\Struct {
    public function __construct(\Kaitai\Struct\Stream $io, \Kaitai\Struct\Struct $parent = null, InstanceStdArray $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }
    private function _parse() {
        $this->ofs = $this->_io->readU4le();
        $this->entrySize = $this->_io->readU4le();
        $this->qtyEntries = $this->_io->readU4le();
    }
    protected $entries;
    public function entries(): array {
        if ($this->entries !== null)
            return $this->entries;
        $_pos = $this->_io->pos();
        $this->_io->seek($this->ofs);
        $this->entries = [];
        $n = $this->qtyEntries();
        for ($i = 0; $i < $n; $i++) {
            $this->entries[] = $this->_io->readBytes($this->entrySize);
        }
        $this->_io->seek($_pos);
        return $this->entries;
    }
    protected $ofs;
    protected $entrySize;
    protected $qtyEntries;
    public function ofs() { return $this->ofs; }
    public function entrySize() { return $this->entrySize; }
    public function qtyEntries() { return $this->qtyEntries; }
}
```
IfStruct:
```php
<?php
// This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild

namespace Kaitai\Struct\Tests;

class IfStruct extends \Kaitai\Struct\Struct {

    public function __construct(\Kaitai\Struct\Stream $io, \Kaitai\Struct\Struct $parent = null, IfStruct $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }

    private function _parse() {
        $this->op1 = new IfStruct\Operation($this->_io, $this, $this->_root);
        $this->op2 = new IfStruct\Operation($this->_io, $this, $this->_root);
        $this->op3 = new IfStruct\Operation($this->_io, $this, $this->_root);
    }

    protected $op1;
    protected $op2;
    protected $op3;

    public function op1() { return $this->op1; }

    public function op2() { return $this->op2; }

    public function op3() { return $this->op3; }
}

namespace Kaitai\Struct\Tests\IfStruct;

class Operation extends \Kaitai\Struct\Struct {

    public function __construct(\Kaitai\Struct\Stream $io, \Kaitai\Struct\Tests\IfStruct $parent = null, \Kaitai\Struct\Tests\IfStruct $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }
    private function _parse() {
        $this->opcode = $this->_io->readU1();
        if ($this->opcode == 84) {
            $this->argTuple = new ArgTuple($this->_io, $this, $this->_root);
        }
        if ($this->opcode == 83) {
            $this->argStr = new ArgStr($this->_io, $this, $this->_root);
        }
    }
    protected $opcode;
    protected $argTuple;
    protected $argStr;
    public function opcode() { return $this->opcode; }
    public function argTuple() { return $this->argTuple; }
    public function argStr() { return $this->argStr; }
}

class ArgTuple extends \Kaitai\Struct\Struct {

    public function __construct(\Kaitai\Struct\Stream $io, Operation $parent = null, \Kaitai\Struct\Tests\IfStruct $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }
    private function _parse() {
        $this->num1 = $this->_io->readU1();
        $this->num2 = $this->_io->readU1();
    }
    protected $num1;
    protected $num2;
    public function num1() { return $this->num1; }
    public function num2() { return $this->num2; }
}
class ArgStr extends \Kaitai\Struct\Struct {

    public function __construct(\Kaitai\Struct\Stream $io, Operation $parent = null, \Kaitai\Struct\Tests\IfStruct $root = null) {
        parent::__construct($io, $parent, $root);
        $this->_parse();
    }
    private function _parse() {
        $this->len = $this->_io->readU1();
        $this->str = $this->_io->readStrByteLimit($this->len, "UTF-8");
    }
    protected $len;
    protected $str;
    public function len() { return $this->len; }
    public function str() { return $this->str; }
}
```
A note: the all `array_push($arr, expr);` calls in the previous generated classes can be replaced with the `$arr[] = expr;` calls, it should be faster.

### DIFF
--- a/spec/php/IfStructTest.php
+++ b/spec/php/IfStructTest.php
@@ -3,19 +3,16 @@ namespace Kaitai\Struct\Tests;
 
 class IfStructTest extends TestCase {
     public function testIfStruct() {
-        $this->markTestIncomplete();
-/*
-        IfStruct r = IfStruct::fromFile(self::SRC_DIR_PATH . "if_struct.bin");
+        $r = IfStruct::fromFile(self::SRC_DIR_PATH . "/if_struct.bin");
 
-        $this->assertEquals(r.op1().opcode(), 0x53);
-        $this->assertEquals(r.op1().argStr().str(), "foo");
+        $this->assertEquals(0x53, $r->op1()->opcode());
+        $this->assertEquals("foo", $r->op1()->argStr()->str());
 
-        $this->assertEquals(r.op2().opcode(), 0x54);
-        $this->assertEquals(r.op2().argTuple().num1(), 0x42);
-        $this->assertEquals(r.op2().argTuple().num2(), 0x43);
+        $this->assertEquals(0x54, $r->op2()->opcode());
+        $this->assertEquals(0x42, $r->op2()->argTuple()->num1());
+        $this->assertEquals(0x43, $r->op2()->argTuple()->num2());
 
-        $this->assertEquals(r.op3().opcode(), 0x53);
-        $this->assertEquals(r.op3().argStr().str(), "bar");
-*/    
+        $this->assertEquals(0x53, $r->op3()->opcode());
+        $this->assertEquals("bar", $r->op3()->argStr()->str());
     }
 }
--- a/spec/php/InstanceIoUserTest.php
+++ b/spec/php/InstanceIoUserTest.php
@@ -3,15 +3,12 @@ namespace Kaitai\Struct\Tests;
 
 class InstanceIoUserTest extends TestCase {
     public function testInstanceIoUser() {
-        $this->markTestIncomplete();
-/*
-        InstanceIoUser r = InstanceIoUser::fromFile(self::SRC_DIR_PATH . "instance_io.bin");
+        $r = InstanceIoUser::fromFile(self::SRC_DIR_PATH . "/instance_io.bin");
 
-        $this->assertEquals(r.qtyEntries(), 3);
+        $this->assertEquals(3, $r->qtyEntries());
 
-        $this->assertEquals(r.entries().get(0).name(), "the");
-        $this->assertEquals(r.entries().get(1).name(), "rainy");
-        $this->assertEquals(r.entries().get(2).name(), "day it is");
-*/    
+        $this->assertEquals("the", $r->entries()[0]->name());
+        $this->assertEquals("rainy", $r->entries()[1]->name());
+        $this->assertEquals("day it is", $r->entries()[2]->name());
     }
 }
--- a/spec/php/InstanceStdArrayTest.php
+++ b/spec/php/InstanceStdArrayTest.php
@@ -3,18 +3,15 @@ namespace Kaitai\Struct\Tests;
 
 class InstanceStdArrayTest extends TestCase {
     public function testInstanceStdArray() {
-        $this->markTestIncomplete();
-/*
-        InstanceStdArray r = InstanceStdArray::fromFile(self::SRC_DIR_PATH . "instance_std_array.bin");
+        $r = InstanceStdArray::fromFile(self::SRC_DIR_PATH . "/instance_std_array.bin");
 
-        $this->assertEquals(r.ofs(), 0x10);
-        $this->assertEquals(r.qtyEntries(), 3);
-        $this->assertEquals(r.entrySize(), 4);
+        $this->assertEquals(0x10, $r->ofs());
+        $this->assertEquals(3, $r->qtyEntries());
+        $this->assertEquals(4, $r->entrySize());
 
-        $this->assertEquals(r.entries().size(), 3);
-        $this->assertEquals(r.entries().get(0), new byte[] { 0x11, 0x11, 0x11, 0x11 });
-        $this->assertEquals(r.entries().get(1), new byte[] { 0x22, 0x22, 0x22, 0x22 });
-        $this->assertEquals(r.entries().get(2), new byte[] { 0x33, 0x33, 0x33, 0x33 });
-*/    
+        $this->assertEquals(3, count($r->entries()));
+        $this->assertEquals("\x11\x11\x11\x11", $r->entries()[0]);
+        $this->assertEquals("\x22\x22\x22\x22", $r->entries()[1]);
+        $this->assertEquals("\x33\x33\x33\x33", $r->entries()[2]);
     }
 }
--- a/spec/php/PositionInSeqTest.php
+++ b/spec/php/PositionInSeqTest.php
@@ -3,16 +3,8 @@ namespace Kaitai\Struct\Tests;
 
 class PositionInSeqTest extends TestCase {
     public function testPositionInSeq() {
-        $this->markTestIncomplete();
-/*
-        PositionInSeq r = PositionInSeq::fromFile(self::SRC_DIR_PATH . "position_in_seq.bin");
+        $r = PositionInSeq::fromFile(self::SRC_DIR_PATH . "/position_in_seq.bin");
 
-        ArrayList<Integer> expected = new ArrayList<Integer>(3);
-        expected.add(1);
-        expected.add(2);
-        expected.add(3);
-
-        $this->assertEquals(r.numbers(), expected);
-*/    
+        $this->assertEquals($r->numbers(), [1, 2, 3]);
     }
 }


GreyCat:
I see that for `repeat` with known number of iterations, you're still using dynamic growth of an array:

```php
        $this->numbers = [];
        $n = $this->header()->qtyNumbers();
        for ($i = 0; $i < $n; $i++) {
            //array_push($this->numbers, $this->_io->readU1());
            $this->numbers[] = $this->_io->readU1();
        }
```

This is probably slow - it's usually faster to pre-allocate array of a certain size that we know beforehand. But, anyway, I'll implement it as you say and we'll do proper benchmarks a bit later.

GreyCat:
>        // PHP does not have a special construct for initialization of the size of an array.
>        // So we use an `if` with throwing of an exception.
>        // Instead of throwing the \OutOfBoundException we could use another approach:
>        // apply the array_slice() and therefore trim an array to the required length,
>        // e.g.: $this->entries = array_slice($this->entries, 0, $this->qtyEntries);
>        // but in this case we will not know that some elements have been trimmed.
>        if (count($this->entries) !== $n) {
>            throw new \OutOfBoundsException();
>        }

Ah, got it. Actually, I don't see much point in checking for that. The loop is guaranteed to execute `$n` times, in all other cases it would throw some sort of exception (i.e. I/O error, EOF, etc). No one else appends to our array, so array is pretty much guaranteed to have exactly `$n` members in result.

GreyCat:
I've also noticed that you write accessing properties as method calls, not as properties, i.e.:

```php
        $n = $this->header()->qtyNumbers();
```

and not

```php
        $n = $this->header->qtyNumbers;
```

What's the rationale behind that? And why doesn't it work (I get "Trying to get property of non-object" error in the latter case)? I thought that they were supposed to be equivalent?

markbook2:
It would be good to discuss it. It is due what is called in PHP [overloading](http://php.net/manual/en/language.oop5.overloading.php). When the `$this->header` property is present/exists the magic `$this->__get()` method will not be called by PHP and therefore the $this->header() method, because the `$this->header` is accessible for reading (visible in the current scope). As result the $this->header will not be initialized and will contain the default value === null here:
```php
        $n = $this->header->qtyNumbers;
```
This is why the error is triggered in such case.

For client code (or tests) the `__get()` method will be called and therefore the respective method because the client code can't read the `protected` or `private` properties (the are not visible in the current scope) - the `__get()` method will not be called for the `public` properties only as they can be read by client.

markbook2:
As addition for the previous note: sometimes it may be a typo or not necessary as here:
```php
class InstanceIoUser extends \Kaitai\Struct\Struct {
// ...
    private function _parse() {
        $this->qtyEntries = $this->_io->readU4le();
        $this->entries = [];//new ArrayList <InstanceIoUser\Entry>((int)($this->qtyEntries));
        $n = $this->qtyEntries(); // not necessary to use the method call here, can be replaced with the property fetch: $this->qtyEntries
// ...
```
as there is no special logic in the qtyEntries() method:
```php
public function qtyEntries() { return $this->qtyEntries; }
```
So the above chunk of the code can be replaced with:
```php
class InstanceIoUser extends \Kaitai\Struct\Struct {
// ...
    private function _parse() {
        $this->qtyEntries = $this->_io->readU4le();
        $this->entries = [];//new ArrayList <InstanceIoUser\Entry>((int)($this->qtyEntries));
        $n = $this->qtyEntries;
// ...
```
***
The general approach may be generalized as: if the method body contains only `return $this->propertyName;` use the property fetch in other case use the method call.

GreyCat:
Ok, I get it. It's generally a bad idea to determine *how* to call - i.e. with `()` or without `()` - by some internal knowledge of class implementation. It breaks one of the basic principles of OOP - [encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)). Adding mechanism to determine whether it's "our own" property or "some external class" property and dealing with them in different manner exclusively for PHP is, of course, possible, but would be really cubersome.

But given that you've said it's only a problem when someone access the class from outside, I see two simple solutions:

1. Just use method access calls `()` for all generated code. It's probably the easiest to implement, and, given that I see that properties actually work with one extra implicit `__get` call and extra hash lookup, it might be even faster.

2. Go with C++ way and add some sort of prefix for protected internal variables storage (for example, like in C++ - it uses `m_foo` instead of `foo`) — and then it would be easy to distinguish between both method calls and properties in `__get` implementation.

GreyCat:
I've tried method (1), i.e. adding `()` everywhere - alas, it doesn't seem to work in *all* places. External stuff (i.e. from the test) accessing like that:

```php
class NavParentTest extends TestCase {
    public function testNavParent() {
        $r = NavParent::fromFile(self::SRC_DIR_PATH . "/nav.bin");

        $this->assertEquals(2, $r->header->qtyEntries);
```

fails on this very line with `Trying to get property of non-object`, probably not even trying to get to generated class internals.

The majority of internal problems are solved, but I'm not sure if that's the right way if external access still doesn't work as intended. In many other tests it works exactly because you've rewrote them to use method access, i.e. in FixedStructTest:

```php
        $this->assertEquals(255, $r->hdr()->uint8());
```

Also, it seems that there's problem accessing special stuff like `_io()` using method getters, i.e. this fails:

```php
$io = $this->_root()->strings()->_io();
```

I notice that you've wrote it as

```php
$io = $this->_root->strings()->_io;
```

in your rendition, but, again, that's exactly what I'd like to avoid.

markbook2:
I agree with replacing the `property fetch` expressions with the `method call` expressions. I have [removed](https://github.com/kaitai-io/kaitai_struct_php_runtime/commit/d85d0dae98f6c9771d444b3e326c03fd468e7bc7) the `__get()` method as it may lead to confusion.

GreyCat:
Um, actually I haven't proposed such a thing. I proposed to rename internal properties to something that won't clash with method names.

markbook2:
As I understood you want to replace the `$r->header->qtyEntries` with the `$r->header()->qtyEntries()`, right?

GreyCat:
Actually, no. My second proposal was:

> Go with C++ way and add some sort of prefix for protected internal variables storage (for example, like in C++ - it uses m_foo instead of foo) — and then it would be easy to distinguish between both method calls and properties in __get implementation.

Ditching and leaving only getter methods whole properties thing is also an option (actually, we've just decided to do exactly that with Perl), but it's a last resort, if everything else fails.

markbook2:
> Go with C++ way and add some sort of prefix for protected internal variables storage (for example, like in C++ - it uses m_foo instead of foo) — and then it would be easy to distinguish between both method calls and properties in __get implementation.

Why do you need to distinguish between method calls and properties? The problem was in the following:
if a property is available the __get() method will not be called and therefore the respective method.

Will adding of prefix help to solve this problem? I think no. The replacing of the property fetch with methods calls will solve it.

GreyCat:
Let me demonstrate what I have in mind:

```php
<?php
class Foo {
    protected $m_bar;

    public function __construct() {
        $this->m_bar = 42;
    }

    public function __get($name) {
        if (method_exists($this, $name)) {
            return $this->$name();
        }
        throw new \RuntimeException("Cannot access the property '" . get_class($this) . '::' . $name . "'");   
    }

    // Regular getter for a sequence attribute
    public function bar(): int {
        return $this->m_bar;
    }

    // Calculated instance, based on sequence attribute 
    public function baz(): int {
        return $this->bar + 42; // note that it is ->bar, not ->m_bar
    }

    // Calculated instance, based on sequence attribute *and* other instance
    public function qux(): int {
        return $this->bar + $this->baz;
    }
}

$foo = new Foo();

// Testing all types of invocations:
print $foo->bar . "\n";
print $foo->bar() . "\n"; 
print $foo->baz . "\n";
print $foo->baz() . "\n";
print $foo->qux . "\n";
print $foo->qux() . "\n";
```

All 6 types work - both with and without `()` and for both "sequence attributes" (i.e. fields parsed in constructor) and instances (i.e. fields that are *not* parsed in constructor, but generated by a method call) in all combinations.

Any problems with this approach?

markbook2:
This approach will work too.

markbook2:
We could add the _io() method to the Struct:
```
    public function _io(): Stream {
        return $this->_io;
    }
```
Does it make sense?

GreyCat:
Yeah, definitely.

PR #22, opened 2016-12-16
#38: Test wrong derivation of _parent in value instances

tschoening:
This is a corresponding test, which should fail during `build-formats` already or simply work.


GreyCat:
Ok, first things first: merging the basic test. Thanks for your work!

PR #23, opened 2016-12-19
#39: Invalid "switch"-statement generated for Java.

tschoening:
In case of multiple boolean operations as condition of a switch statement, parenthesis weren't properly generated. This at least in Java created code which didn't compile.

https://github.com/kaitai-io/kaitai_struct_compiler/issues/39

### DIFF
--- a/formats/switch_multi_bool_ops.ksy
+++ b/formats/switch_multi_bool_ops.ksy
@@ -1,0 +1,22 @@
+meta:
+  id: switch_multi_bool_ops
+  endian: le
+seq:
+  - id: opcodes
+    type: opcode
+    repeat: eos
+types:
+  opcode:
+    seq:
+      - id: code
+        type: u1
+      - id: body
+        type:
+          switch-on: "(  (code >   0) and
+                         (code <=  8) and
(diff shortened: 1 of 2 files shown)


GreyCat:
Sorry for late reviews and replies.

This is somewhat border case, but just wanted you to know that we actually have another mechanism to testing expressions. That is [ExpressionSpec](https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/jvm/src/test/scala/io/kaitai/struct/exprlang/ExpressionsSpec.scala) (used for testing parsing of expression string → AST (abstract syntax tree)) and [TranslatorSpec](https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala) (used to test translators' functionality, i.e. AST → expression in target languages).

Both are written using scalatest, which is very close to normal Java test frameworks (i.e. JUnit or TestNG). I'll probably implement this test (or something close) in TranslatorSpec as well. They're much easier to write (i.e. you don't need to create separate .ksy file and "useless" specs in target languages, as we only want to test translation of a single expression).

PR #24, opened 2016-12-20
Added a test to map value instances to an enum.

tschoening:
I made a mistake in my KSY file, mapping a `value` instance already containing `enum` instances in a boolean operation/`Ast.expr.IfExp` to an `enum`: I used the keyword `enum` in this case, because I somehow thought that I need to provide the resulting data type for the value instance and that didn't work, because KS mapped the expression to `EnumById`. But there was no ID to map of course... I simply needed to not use the `enum` keyword in this case, but it took me quite some time to understand this.

I've added this test to make the difference easier to remember for me or simply have some example I can search for next time. Maybe you find it a bit useful for others and hope you're are merging it.

### DIFF
--- a/formats/enum_of_value_inst.ksy
+++ b/formats/enum_of_value_inst.ksy
@@ -1,0 +1,21 @@
+meta:
+  id: enum_of_value_inst
+  endian: le
+seq:
+  - id: pet_1
+    type: u4
+    enum: animal
+  - id: pet_2
+    type: u4
+    enum: animal
+enums:
+  animal:
+    4: dog
+    7: cat
+    12: chicken
+instances:
+  pet_3:
+    value: "pet_1 == animal::cat ? 4 : 12"
+    enum: animal
+  pet_4:
+    value: "pet_1 == animal::cat ? animal::dog : animal::chicken"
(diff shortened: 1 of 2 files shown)


GreyCat:
Sure :) Could you also push a "wrong" version into [kaitai_struct_tests/formats_err](https://github.com/kaitai-io/kaitai_struct_tests/tree/master/formats_err)? There are some .ksy files that are meant to *not* compile there already, and I hope some day we'll have proper coverage of all error cases.

For example, it this use-case that you've provided, compiler actually knows that value instances' derived result type is EnumType already, so it should report an error when someone's trying to apply extra `enum` on top of it.

tschoening:
Added such an error test to the same branch, so you can just pull again.

Besides answering the wrong KSY with an error, `InstanceSpec` could be made compatible by not always mapping to `EnumById`. That's what I implemented first until I recognized that the additional `enum` was wrong. Just an idea, both approaches have Pros and Cons to me...

        // Wrap everything in EnumById if "enum" is used and "value" contains an expression using
        // integers, but take care of exceptions like "EnumByLabel" in some situations as well.
        val value2 = ParseUtils.getOptValueStr(srcMap, "enum", path) match {
          case None =>
            value
          case Some(enumName) =>
            value match {
              case Ast.expr.IfExp(_, ifTrue: Ast.expr, _) =>
                ifTrue match {
                  case Ast.expr.EnumByLabel(_,_) => value
                  case _ => Ast.expr.EnumById(Ast.identifier(enumName), value)
                }
              case _ => Ast.expr.EnumById(Ast.identifier(enumName), value)
            }

PR #26, opened 2017-01-02
added switch_manual_str_else test

adamiwaniuk:


GreyCat:
Thanks! I'll port it to other languages as well, I suspect there might be problems with Python and Perl here...

PR #32, opened 2017-03-01
Added markTestAsNotRisky() and marked some tests as skipped and some …

markbook2:
Added markTestAsNotRisky() and marked some tests as skipped and some as not risky


GreyCat:
Thanks! Top notch work, as always :)

PR #33, opened 2017-07-01
Added process_coerce_switch test + data

mickdekkers:
@GreyCat here you go :)

Unfortunately I couldn't get the dev setup to work so I haven't been able to port or test it yet.


GreyCat:
Super! Thanks! That's a lot of work & help :)

PR #34, opened 2017-08-12
Ported a few more tests to Go

nolta:


GreyCat:
Thanks again! You seem to be doing quite a lot of progress with Go port — I believe we should make some notes about that in https://github.com/kaitai-io/kaitai_struct/issues/146 :)

GreyCat:
A few notes on your conversions — it looks like you haven't tested it (as compiler was obviously broken, I've tried to fix it now), and it caught my attention that conversion script ports assertions in the wrong order. Technically, it's `assert.Equal(t, expected, actual)`, i.e. something like

```go
assert.Equal(t, byte(0x50), h.One)                                                                                                        
```

However, your ports use "actual, expected" order:

```go
assert.EqualValues(t, r.StrPad, "str1")
```

This leads to misleading error messages, i.e:

```
Got error:
=== RUN   TestBytesPadTerm
--- FAIL: TestBytesPadTerm (0.00s)
        Error Trace:    bytes_pad_term_test.go:26
        Error:          Not equal: 
                        expected: []uint8([]byte{0x73, 0x74, 0x72, 0x31, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40, 0x40})
                        received: string("str1")
```

In fact, it's the exact opposite, I guess: the proper answer is specified as a string (and that's an error in the spec), but generated code uses byte array.

nolta:
I did indeed run the tests, against the compiler in the "go" branch.

PR #37, opened 2017-08-24
Added tests for Lua

adrianherrera:
Hi there!

I wrote a Lua backend for Kaitai Struct. These are the tests. They require luaunit to run (https://github.com/bluebird75/luaunit). The runtime is available at https://github.com/adrianherrera/kaitai_struct_lua_runtime


GreyCat:
Just tested it, got:

```
Failed tests:
-------------
1) TestFixedStruct.test_fixed_struct
spec/lua/test_fixed_struct.lua:13: expected: 1.844674407371e+19, actual: -1
stack traceback:
        spec/lua/test_fixed_struct.lua:13: in upvalue 'TestFixedStruct.test_fixed_struct'

Ran 125 tests in 0.016 seconds, 124 successes, 1 failure
```

A very impressive result indeed!

I'll send you invitation to join this repository, so you can edit these files right away, without pull requests, if you'd prefer to :)

adrianherrera:
Ah damn, how did that happen, I swear I got them all working. I'll fix it up tomorrow!

GreyCat:
Another quick question: could you change `ci-lua` script to make it save test run results as some sort of file (preferably, something close to JUnit xml - from a quick peek at luaunit I believe it supports that), located somewhere in `test_out/lua/...`?

adrianherrera:
So I get 100% success rate on my tests:

```
=========================================================
Ran 125 tests in 0.005 seconds, 125 successes, 0 failures
OK
```

I'm not sure why that one is failing on yours!

GreyCat:
Yeah, this is strange. I've added Lua test running in our Travis CI, and it got 125/125 there as well.

GreyCat:
I've just added Lua CI to aggregation, so finally Lua is a first class citizen now — http://kaitai.io/ci/ — congratulations (and it already surpassed Perl) :)

GreyCat:
Meanwhile, I also have a few suggestions / questions about Lua:

* It's generally not a very good idea to have a list of tests to be invoked in some kind of file, i.e. [that one](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/spec/lua/run_test_suite.lua#L6-L134). It's hard to maintain, it's one extra commit for every added tests, etc. Is it somehow possible to read directory listing and `require(...)` every test spec found automatically?
* As KS continues to evolve, tests are going to be added. Typically, we use (semi-)automated test converters to translate existings specs from one language into anothers — that's [spec_ruby_to_cpp_stl](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/spec_ruby_to_cpp_stl), [spec_ruby_to_csharp](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/spec_ruby_to_csharp), [spec_ruby_to_go](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/spec_ruby_to_go), etc. You see the general pattern, in most cases, the "original" test spec is written in Ruby and then converted from there. In many cases, translation is fully automatic, I just add one spec for Ruby and then launch [spec_translate_missing](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/spec_translate_missing) which does all the conversions. Would it be possible to create similar tool for Lua? I bet you weren't manually rewriting all these 125 specs, so you've probably already have some utilities?

GreyCat:
* As you might have seen in the [CI](http://kaitai.io/ci), Lua support currently lacks "expression translation tests". These are generally located in [jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec](https://github.com/kaitai-io/kaitai_struct_compiler/blob/master/jvm/src/test/scala/io/kaitai/struct/translators/TranslatorSpec.scala). The first step would be to add LuaTranslator into that at all — chances are it will automatically pass like half of the trivial tests or more (it will catch up when test condition is using "for all languages" or "for all languages except for X"). The rest is probably need to be added.

GreyCat:
And yet another quick question: we were starting [Lua-based Wireshark dissectors initiative](https://github.com/kaitai-io/kaitai_struct/issues/50). Could you estimate if, by any chance, your Lua support will help to resurrect that Wireshark dissectors support?

adrianherrera:
- Yeah, I hate having `run_test_suite.lua`, but I couldn't find away around it. That said, I didn't look too much into it. I'll have a look and try to remove this.
- Tbh I didn't even notice those converters! Yeah, I'll add something for Lua
- Again, I didn't notice the "expression translation tests" (sorry!!). I'll look into it
- Hmm interesting, I have no idea about Wireshark dissectors (my interest is in S2E, which also uses Lua as a scripting language)

adrianherrera:
Expression translation tests added!

GreyCat:
Thanks!

adrianherrera:
And now I've added the test converter for Lua as well!

I've asked the luaunit devs if there is a "test discovery" mode similar to Python's unittest module, but for now I don't think that there's any way to remove the test runner.

koczkatamas:
I think the important thing here is to remove those require lines from the test runner, not removing the test runner itself.

I presume we can do that by listing all the files in the current directory, filter them by some logic (begins with `test_`?) and call require on them.

So this way we don't have to modify the test runner the next time, just add a new test file into this folder (this can even happen automatically from a ruby->lua translator script and not manually) and it will run this test too automatically when we call the test runner.

PR #38, opened 2017-08-25
Configured junit output for CI Lua tests

adrianherrera:
I couldn't push directly to the repo (permission error), so I've made a PR instead.


GreyCat:
Thanks! I'll try to fix permissions problem ASAP.

PR #39, opened 2017-08-25
Fixed some Lua tests so they actually run

adrianherrera:
Still can't push directly to kaitai_struct_tests :)

I had a look at the CI server and fixed two of the tests so that they run.


GreyCat:
Thanks! I've just fixed the permissions, should work now.

If you want to test CI, please pull these commits (i.e. update the pointer for `tests/` submodule) in the [main umbrella project](https://github.com/kaitai-io/kaitai_struct/) and push. It should launch the CI process, and in 8-9 minutes you'll see http://kaitai.io/ci/ updated.

PR #40, opened 2018-03-30
Add golang enum test

cugu:


GreyCat:
Thanks!

GreyCat:
JFYI: I ended up disabling enum_0 test for now, as it breaks compilation and thus whole test run. Unfortunately, our build script can only skip compilation errors in compiled format files, but not those that happen in spec files.

PR #43, opened 2018-08-09
Partial implementation of Rust output

CWood1:
Created a set of tests for Rust output, as discussed in https://github.com/kaitai-io/kaitai_struct/issues/22.


ci-rust
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+. ./config
+
+mkdir -p "$TEST_OUT_DIR/rust"
+
+pushd spec/rust
+cargo test-junit \
+      --name "../../$TEST_OUT_DIR/rust/results.xml"
+popd
```
GreyCat: You can actually just do `cd spec/rust` and not worry about returning back — this one is always executed in a separate shell, not sourced or anything like that.

GreyCat:
Enormous thanks for your contribution, @CWood1!

First things first: could you remove `spec/ks/out` generated files, as they should not be checked in and should not really be part of any commit. Moreover, having >1K of changed files makes it hard to review this one.

CWood1:
Sorry, didn't realise I'd committed the autogenned files. My bad. I'll get things cleaned up shortly, and back to you.

CWood1:
I've fixed that which you told me, hopefully this should do the trick. Thanks for the feedback, and let me know if there's anything more I need to look at

GreyCat:
I see that you've removed `spec/rust` autogenerated files as well. That would mean that these files are always expected to be autogenerated in all cases for Rust? It's not the case (at least not yet) for all other languages, as we still have quite a few non-autogenerated specs. And, more than likely, stuff like `debug*` or exception handling will stay language-specific for foreseeable time.

CWood1:
`build.rs` actually copies the files under test into `spec/rust/src`, and generates `lib.rs` from this. It doesn't generate them itself, so they'd need to exist beforehand, but yes, they're expected to be autogenned.

The tests are still manually copied, however, though I would personally make an argument to automate this step as well.

CWood1:
I've just tried uninstalling and reinstalling it, and I get the same thing. Strangely, if I build from the repo, it works. I've filed an issue with the maintainers (https://github.com/rustation/cargo-test-junit/issues/8), though I'll admit to not really knowing what's causing this right now

GreyCat:
Could you outline for me how to "build it from repo"? I've tried cloning their git repo, doing `cargo build` succeeds, but afterwards `ci-rust` still complains that command is missing. I guess I need to install it somehow?

GreyCat:
I think I've figured it out, it's `cargo install --path .` in cloned repo.

GreyCat:
Ok, next stop: `ci-rust` results in compilation failure, dumped into results.xml as:

```
can&apos;t find crate for `rust`
 --&gt; tests/test_expr_array.rs:4:1
  |
4 | extern crate rust;
  | ^^^^^^^^^^^^^^^^^^ can&apos;t find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: Could not compile `rust`.
warning: build failed, waiting for other jobs to finish...
error[E0463]: can&apos;t find crate for `rust`
 --&gt; tests/test_nav_parent3.rs:4:1
  |
4 | extern crate rust;
  | ^^^^^^^^^^^^^^^^^^ can&apos;t find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error[E0463]: can&apos;t find crate for `rust`
 --&gt; tests/test_instance_std.rs:4:1
  |
4 | extern crate rust;
  | ^^^^^^^^^^^^^^^^^^ can&apos;t find crate

error[E0463]: can&apos;t find crate for `rust`
 --&gt; tests/test_process_rotate.rs:4:1
  |
4 | extern crate rust;
  | ^^^^^^^^^^^^^^^^^^ can&apos;t find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: Could not compile `rust`.
warning: build failed, waiting for other jobs to finish...
error: Could not compile `rust`.
warning: build failed, waiting for other jobs to finish...
error: Could not compile `rust`.

To learn more, run the command again with --verbose.
```

Any ideas what can I fix to make it compile?

CWood1:
That's frustrating. I just got the same error, but running again fixed it. That leads me to believe `cargo` attempts to first compile the tests, then builds the project with `build.rs`. Stopgap solution is to run the command twice, but something more robust is definitely needed if this is going into CI

GreyCat:
I've tried running `ci-rust` multiple times, but it turns out that dumps complaints about random 4 files every time (probably some sort of an unstable sort or race?). Never got it run a successful test yet :(

CWood1:
Actually, what am I saying? There is a robust solution, if this is legitimately the issue. Run `cargo build` first, then `cargo test` immediately afterwards

GreyCat:
Running `cargo build` in `spec/rust` succeds, but `cargo test` afterwards in `spec/rust` actually ends in more or less the same manner:

```
error[E0463]: can't find crate for `rust`
 --> tests/test_switch_manual_int_else.rs:4:1
  |
4 | extern crate rust;
  | ^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: Could not compile `rust`.
warning: build failed, waiting for other jobs to finish...
error[E0463]: can't find crate for `rust`
 --> tests/test_instance_user_array.rs:4:1
  |
4 | extern crate rust;
  | ^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error[E0463]: can't find crate for `rust`
 --> tests/test_index_to_param_eos.rs:4:1
  |
4 | extern crate rust;
  | ^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error[E0463]: can't find crate for `rust`
 --> tests/test_process_xor4_const.rs:4:1
  |
4 | extern crate rust;
  | ^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: Could not compile `rust`.
warning: build failed, waiting for other jobs to finish...
error: Could not compile `rust`.
warning: build failed, waiting for other jobs to finish...
error: Could not compile `rust`.
```

CWood1:
What's the directory listing of `spec/rust/src`? I've tried a few times now, and it seems to be intermittent whether `cargo build && cargo test` actually succeeds or not

GreyCat:
There's no such directory, only `tests` and `target` (which should be either listed in gitignore too, or, better yet, moved to `compiled/rust/target` or something like that)

CWood1:
Okay, I've found the issue, I believe. If I'm correct in this, a temporary solution would be to `mkdir src/`, then `cargo build && cargo test`. I'll continue to look into why this only works some of the time.

Going forwards, `build.rs` needs updating to first create `src/`, then to copy everything in.

`target/` can be safely gitignored. `tests`, on the other hand, would need more infrastructure first. These are generated by the test translator, but have been manually copied over. I agree, however, that they should be copied automatically long term. I've added this to my list of things to do.

GreyCat:
Ok, I've got to try that, and there is some progress. After manual creation of `src/`, compilaion (`cargo build`) started to deliver another set of errors, this time related to `src/*` contents:

```
error[E0432]: unresolved import `imported_2`
  --> src/imported_1.rs:11:5
   |
11 | use imported_2::Imported2;
   |     ^^^^^^^^^^ Maybe a missing `extern crate imported_2;`?

error[E0432]: unresolved import `hello_world`
  --> src/imports0.rs:11:5
   |
11 | use hello_world::HelloWorld;
   |     ^^^^^^^^^^^ Maybe a missing `extern crate hello_world;`?

error[E0432]: unresolved import `imported_root`
  --> src/imported_and_rel.rs:11:5
   |
11 | use imported_root::ImportedRoot;
   |     ^^^^^^^^^^^^^ Maybe a missing `extern crate imported_root;`?

error[E0432]: unresolved import `hello_world`
  --> src/cast_to_imported.rs:11:5
   |
11 | use hello_world::HelloWorld;
   |     ^^^^^^^^^^^ Maybe a missing `extern crate hello_world;`?

error[E0425]: cannot find value `_root` in this scope
  --> src/nav_parent3.rs:55:77
   |
55 |             self.tags.push(Box::new(NavParent3__Tag::new(self.stream, self, _root)?));
   |                                                                             ^^^^^ not found in this scope

error[E0425]: cannot find value `_root` in this scope
   --> src/nav_parent3.rs:107:93
    |
107 |                 self.tagContent = Box::new(NavParent3__Tag__TagChar::new(self.stream, self, _root)?);
    |                                                                                             ^^^^^ not found in this scope

error[E0425]: cannot find value `_root` in this scope
  --> src/position_to_end.rs:49:79
   |
49 |         self.index = Box::new(PositionToEnd__IndexObj::new(self.stream, self, _root)?);
   |                                                                               ^^^^^ not found in this scope

error[E0425]: cannot find value `_root` in this scope
  --> src/fixed_struct.rs:49:73
   |
49 |         self.hdr = Box::new(FixedStruct__Header::new(self.stream, self, _root)?);
   |                                                                         ^^^^^ not found in this scope

warning: unused import: `std::io::Cursor`
 --> src/imported_1.rs:6:5
  |
6 | use std::io::Cursor;
  |     ^^^^^^^^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

warning: unused import: `std::vec::Vec`
 --> src/imported_1.rs:7:5
  |
7 | use std::vec::Vec;
  |     ^^^^^^^^^^^^^
warning: unused import: `std::io::Cursor`
 --> src/opaque_external_type_02_child.rs:6:5
  |
6 | use std::io::Cursor;
  |     ^^^^^^^^^^^^^^^

warning: unused import: `std::vec::Vec`
 --> src/opaque_external_type_02_child.rs:7:5
  |
7 | use std::vec::Vec;
  |     ^^^^^^^^^^^^^

warning: unused import: `std::io::Cursor`
 --> src/nav_root.rs:6:5
  |
6 | use std::io::Cursor;
  |     ^^^^^^^^^^^^^^^

warning: unused import: `std::io::Cursor`
 --> src/nav_parent_false.rs:6:5
  |
6 | use std::io::Cursor;
  |     ^^^^^^^^^^^^^^^

warning: unused import: `std::io::Cursor`
 --> src/default_endian_expr_exception.rs:6:5
  |
6 | use std::io::Cursor;
  |     ^^^^^^^^^^^^^^^

```
etc. That's a pretty long list that ends with:

```
error[E0308]: mismatched types
  --> src/expr_1.rs:53:16
   |
53 |         return self.lenOf1Mod;
   |                ^^^^^^^^^^^^^^ expected i32, found enum `std::option::Option`
   |
   = note: expected type `i32`
              found type `std::option::Option<i32>`

error[E0308]: mismatched types
  --> src/expr_1.rs:60:24
   |
60 |         self.str1Len = self.str1.len();
   |                        ^^^^^^^^^^^^^^^ expected enum `std::option::Option`, found usize
   |
   = note: expected type `std::option::Option<i32>`
              found type `usize`

error[E0308]: mismatched types
  --> src/expr_1.rs:61:16
   |
61 |         return self.str1Len;
   |                ^^^^^^^^^^^^ expected i32, found enum `std::option::Option`
   |
   = note: expected type `i32`
              found type `std::option::Option<i32>`

error: aborting due to 997 previous errors

Some errors occurred: E0277, E0308, E0317, E0369, E0425, E0432, E0599, E0609.
For more information about an error, try `rustc --explain E0277`.
```

May be there's some problem with the Rust version I'm using (1.28.0)?

GreyCat:
Could you also explain the role of `build.rs`? From what I see now, it does what could be done in 2 lines of shell script, i.e.:

```shell
rm -rf spec/rust/src
mkdir -p spec/rust/src
cp -r compiled/rust/* spec/rust/src
```

Current implementation actually forgets to do that `mkdir` and, what's more important, it screws up compilation flow, as you can't guarantee that `cargo test` process would pick up these newly copied in files, as `cargo test` invocation happened before these files appeared.

Given that we use `ci-rust` shell script to launch it anyway, why don't we:

1. Do that file copying in `ci-rust`?
2. Use symlink rather than copying (`ci-go` actually uses the very similar technique, but for bringing in latest go runtime without fetching github repo directly)?

CWood1:
Regarding the errors, that's what I want right now. Those are "this is a partial implementation and there's more work to do" errors, rather than "there's problems with the test harness" errors. Each different type of error has a todo item somewhere in my workflow, and as soon as we're both satisfied this PR is complete, that's what I intend to go back to working on.

Regarding `build.rs`, it's also responsible for building `src/lib.rs`. This could also be done in `ci-rust` I suppose, but it would need one line per copied source file, meaning the overall complexity of the Bash goes up. Personally, I did it this way because I know Rust better than Bash, though I'll happily rewrite `ci-rust` to do away with `build.rs` if you think it's a worthwhile endeavour to do so.

GreyCat:
Ok, so, in other words, it is expected to fail with all these errors now, and probably what you do is that you manually delete these erroneous compilation results / make them void in some other way, to still get results from the good part.

If yes, then actually all compiled languages suffer from the same problem, and there has been several solutions we've been applying. For example:

* `run-java` tries to run compilation of "everything" first, and then, if it fails, it invokes compiler on each generated source file manually. This way we still get `.class` files for everything that passed compilation stage and clear error messages in build log for those who have failed. After we've got all `.class` files we could, we proceed with test run. Due to the way Java handles class loading, tests for missing `.class` files would be just ignored. Also, thanks to us having data which tests failed to compile, we log that into a separate file and show it as special test result in [CI](http://kaitai.io/ci)

* `run-cpp_stl` basically avoids all this by having a project file (`CMakeLists.txt`) which lists which files are known to be "good" and not break the compilation. If "known to be good" file turns out "bad", it still breaks whole C++ target compilation. Not a very good solution, but minimally working and probably the simplest one.

* `run-csharp` takes a very aggressive approach, by doing (1) initial run, just as Java, (2) takes all error messages in the log and just overwrites all files which were causing it with empty files, (3) tries to compile for the second time. Again, not a very good solution (for example, it would hurt `spec/csharp` files, which are not exactly autogenerated and should never be broken), but it works.

* `ci-go` is probably the most complex and the cleanest one. It's the same idea as `run-csharp`, i.e. it removes offending files, but (1) it does so in a loop, as golang compiler only shows a few (~10) error messages on every invocation, not all of them, (2) it deletes only files in `compiled/go`, i.e. ksc results, never touching any `spec/go` files, which are supposed to be manually-written and are always known to be good.

I believe we can apply the same approach as with C# and Go here, i.e. automatically deleting bad files and retrying. I'll give it a try.

CWood1:
Perhaps it would be worth instead, attempting to separate out the failing files, in such a way that no information on the failure is lost but it is all separated in different test cases.

That way, CI can clearly denote "these files don't pass because the code generation is broken", I as a developer can still see what's broken, but the fidelity is a little higher than cargo saying "no" and that's it.

I'm not sure if there's an elegant way to do this.

PR #45, opened 2018-11-19
Fix Python 3 tests

cherue:
Currently Python 3.4.3 is used, because `python3` points to it. This causes all Python 3 tests to not run.

Travis log:
```
$ python3 --version
Python 3.4.3
```
https://github.com/kaitai-io/kaitai_struct/blob/master/.travis.yml#L55 should be changed to `python3.6` too.

### DIFF
--- a/ci-construct3
+++ b/ci-construct3
@@ -3,4 +3,4 @@
 . ./config
 
 rm -rf "$TEST_OUT_DIR/construct3"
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks!

Please note, however, that hopefully soon that will be deprecated, and we'll switch to using new CI architecture with all per-language tests ran in individual environments: https://travis-ci.org/kaitai-io/ci_targets

Both Python 2.7 and Python 3.6 are currently tested there.

PR #47, opened 2019-03-23
Add Go test panic handling defer func()

13rac1:
A panic() called in the tests breaks the test result processor, so Don't Panic and print a friendly stacktrace.

Change made with `sed`:
```bash
sed 's/import (/import (\n\t"runtime\/debug"/g' -i *
sed 's/(t \*testing\.T) {/(t \*testing\.T) {\n\tdefer func() {\n\t\tif r := recover(); r != nil {\n\t\t\tdebug.PrintStack()\n\t\t\tt.Fatal("unexpected panic:", r)\n\t\t}\n\t}()/g' -i *
```

### DIFF
--- a/spec/go/bcd_user_type_be_test.go
+++ b/spec/go/bcd_user_type_be_test.go
@@ -3,6 +3,7 @@
 package spec
 
 import (
+	"runtime/debug"
 	"os"
 	"testing"
 	"github.com/kaitai-io/kaitai_struct_go_runtime/kaitai"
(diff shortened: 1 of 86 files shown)


GreyCat:
Thanks! Probably this requires changes in kst translator as well, i.e. around [here](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/translator/src/main/scala/io/kaitai/struct/testtranslator/specgenerators/GoSG.scala#L47).

PR #48, opened 2019-03-23
Add Go test panic handling defer func() to spec generator

13rac1:
Create a handlePanic() func to create code to catch/handle panics. A separate function for better documentation.

Follow up to #47


13rac1:
No response in over two months. Closing.

PR #49, opened 2019-04-04
Allow "import-path" more than once to not necessarily use OS-specific…

ams-tschoening:
… path separator. This fixes https://github.com/kaitai-io/kaitai_struct/issues/557.


GreyCat:
Thanks!

PR #50, opened 2019-04-04
Adding some test for issue #552.

ams-tschoening:
This adds a test reproducing the problem I encountered in https://github.com/kaitai-io/kaitai_struct/issues/552 in Ruby. Manually changing the generated Ruby makes the test succeed as well, that is where the compiler comes into play.


GreyCat:
Thanks, merging in!

PR #52, opened 2019-04-19
Bring into line with current state of work, to integrate with CI

CWood1:


spec/rust/err
```diff
@@ -0,0 +1,1056 @@
+execve("/nix/store/9bmfacva2jsr3k0ww3wqlcgwnhadqvd6-cargo-1.27.0/bin/cargo", ["cargo", "junit", "--name", "/home/venos/output.xml"], 0x7ffef70f6628 /* 169 vars */) = 0
```
GreyCat: I heavily suspect that this does not belong in the repo.

spec/rust/src/bcd_user_type_be.rs
```diff
@@ -0,0 +1,440 @@
+// This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild
```
GreyCat: These should be in `compiled/rust`, not here.

GreyCat:
Looks like there's a batch of extra files were caught up in this PR. Please remove them, so it will be possible to review it.

PR #54, opened 2019-05-29
Add specification for floating point behavior

bspeice:
The spec currently doesn't make any claim on how NaN's behave, and implementations vary widely (Python doesn't parse NaN at all from `f4` types, C++ ignores trailing `NaN` but not leading `NaN`, etc.). 

This formalizes expectations that `NaN` values are ignored, but I'm certainly open to changing the intended behavior.


GreyCat:
Several concerns on this one:

1. My main concern with this is that actually non-finite values are pretty diverse. Even for NaNs you have [quiet and signaling NaNs](https://en.wikipedia.org/wiki/NaN), and the latter are supposed to trigger an exception / fatal when received in some languages.

There's also infinities and some extra cases like underflows. Shall we test (and expect well-defined behavior) for all of them?

2. For the implementation: we'd probably want to have both min/max on arrays and regular comparisons tested?

3. Would we want to have operations like `1/0.0` and `0/0.0` tested and behaving equally in all languages?

bspeice:
I've personally yet to see a language distinguish between quiet/signaling NaN for purposes of exception/fatal; in my experience you can "catch" the signaling variant by trapping SIGFPE, but I'm not familiar with any modern languages making a functional distinction (C++ has `std::numeric_limits::quiet_NaN` and `std::numeric_limits::signaling_NaN` for different bitwise representations, but they don't function any differently). If there's context I'm missing here, please let me know.

You're correct on the regular comparisons part; adding checks for `<`/`>` would definitely be needed as part of this. That said, infinites compare easily; max over something containing `inf` or min over `-inf` is well-behaved.

To your last point, I don't think we *can* expect similar behavior for these cases across all languages, but you're correct to point out that it may be an issue. I would personally want to treat both `1/0` and `0/0` as undefined behavior to allow for platform-specific idioms (Python treats them both as `ZeroDivisionError`, C++ allows `0/0` as `NaN`), but there's an argument to be made for standardizing. 

Finally, it may be worth skipping specification and just documenting "comparisons and aggregates involving NaN are undefined behavior". I don't have strong preferences one way or the other, just thought this was something left unspecified that may be a good idea to document.

PR #55, opened 2019-09-10
Add Go tests that will work after kaitai-io/kaitai_struct_compiler#169 is merged.

jchv:
When kaitai-io/kaitai_struct_compiler#169 is merged, new tests will compile and pass successfully. This PR contains those tests.

More tests would likely compile and pass if the translator was capable of translating them.

This should not be merged until kaitai-io/kaitai_struct_compiler#169 is, as it will otherwise cause Go tests to fail completely.


jchv:
(Sigh, pointed to wrong PR; will fix this in the commit later when I get home.)

GreyCat:
Merging tests as well, thanks!

PR #56, opened 2019-09-14
Nim is ready for CI

sealmove:
`ci-nim` produces a correct `ci.json` in `$TEST_OUT_DIR/nim`.
This PR goes together with [this one](https://github.com/kaitai-io/kaitai_struct_compiler/pull/170).
Can we merge both so I can proceed with setting up Nim for Travis in [ci_targets repo](https://github.com/kaitai-io/ci_targets)?

### DIFF
--- a/run-nim
+++ b/run-nim
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 . ./config
 
@@ -16,11 +16,9 @@ fi
 rm -rf "$TEST_OUT_DIR/nim"
 mkdir -p "$TEST_OUT_DIR/nim/bin"
 
-i=0 args=()
+args=()
 for var in "$@"
 do
-	args[$i]="Kaitai Struct Compiler Test Suite::Nim: $var"
-	((++i))
+	args="$args \"Kaitai Struct Compiler Test Suite::Nim: $var\""
 done
-
-nim c -r --outdir:"$TEST_OUT_DIR/nim/bin" "spec/nim/test" "${args[@]}"
+eval nim c -r --outdir:"$TEST_OUT_DIR/nim/bin" "spec/nim/test" "$args"
--- a/run-nim
+++ b/run-nim
@@ -1,16 +1,16 @@
-#!/bin/sh
+#!/bin/bash
 
 . ./config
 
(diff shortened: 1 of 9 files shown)


run-nim
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+. ./config
+
+if [ ! -f "spec/nim/test.nim" ]; then
+  RED='\033[1;31m'
```
GreyCat: The rest of these shell scripts actually use tabs for indentations, can I ask you to replace spaces here with tabs for consistency?

run-nim
```diff
@@ -0,0 +1,27 @@
...
+mkdir -p "$TEST_OUT_DIR/nim/bin"
+
+i=0 args=()
+for var in "$@"
+do
+  args[$i]="Kaitai Struct Compiler Test Suite::Nim: $var"
+  ((++i))
+done
+
+echo $arguments
```
GreyCat: There is no `$arguments` variable in this script, only `$args`, am I missing something?

run-nim
```diff
@@ -0,0 +1,27 @@
...
+  exit
+fi
+
+rm -rf "$TEST_OUT_DIR/nim"
+mkdir -p "$TEST_OUT_DIR/nim/bin"
+
+i=0 args=()
+for var in "$@"
+do
+  args[$i]="Kaitai Struct Compiler Test Suite::Nim: $var"
```
GreyCat: These are arrays, they won't work in /bin/sh. Either use /bin/bash, or use a more shell-like approach, such as processing arguments as individual lines with `sed 's/^/Kaitai Struct Compiler Test Suite::Nim: /'`

run-nim
sealmove: Sure, my bad for not noticing.

run-nim
sealmove: Leftover from testing the script, I'll remove it.

run-nim
sealmove: I changed the shebang to /bin/bash for now, but if /bin/sh is preferred feel free to me.

GreyCat:
LGTM, thanks, merging in!

PR #58, opened 2019-09-24
Add new tests to Go compiler for endian switches.

jchv:
This adds new Go tests that will pass after https://github.com/kaitai-io/kaitai_struct_compiler/pull/172 is merged.

PR #59, opened 2019-09-24
GoSG: Support expecting an error.

jchv:
This PR enables generating Go tests that expect an error. This depends on the error type code added to GoCompiler in https://github.com/kaitai-io/kaitai_struct_compiler/pull/173 and the error types added to the Go runtime in https://github.com/kaitai-io/kaitai_struct_go_runtime/pull/17. Additionally, it adds new kst tests that now compile and run (but not the ones in https://github.com/kaitai-io/kaitai_struct_tests/pull/58)

Before:

```
SUMMARY: {"kst"=>82, "passed"=>76, "failed"=>14}
```

After:

```
SUMMARY: {"kst"=>92, "passed"=>81, "failed"=>19}
```

(The number of passed should increase to 84 once #58 is merged.)

There are still a couple of snags in the Go kst compiler, but now around 120 kst files at least compile. (All 120 cannot be included because many depend on kaitai structs that aren't working yet, and unfortunately in Go this catastrophically fails tests. :()


GreyCat:
Thanks again, merging in!

PR #63, opened 2019-10-20
Add tests for EOS exceptions

generalmimon:
I noticed that the JavaScript runtime does not throw an exception when trying to read beyond the end of substream, it just passes. This behavior can cause some errors which are hard to find.
So why not test it in all languages?


GreyCat:
Makes sense, thanks!

I was actually just about to revamp our end of stream error handling, so we'll eventually get these expressed in KST too :)

PR #51, opened 2019-04-08
CalcArrayType needs to be handled like ArrayType at some places…

ams-tschoening:
…or things might break e.g. in case of value instances returning lists of user types.

This is related to https://github.com/kaitai-io/kaitai_struct/issues/561.

PR #64, opened 2019-11-07
Make the builder more compatible with Windows.

ams-tschoening:
During testing updated PRs https://github.com/kaitai-io/kaitai_struct_compiler/pull/160 and https://github.com/kaitai-io/kaitai_struct_tests/pull/51 I recognized build errors related to hard-coded command lines and stuff like that. Changed that to OS-abstraction provided by ruby and the tests execute now.

Various things needed to be changed:

* ":" -> ";" using FIlE::PATH_SEPERATOR
* "/" -> "\" using File.join because some paths simply got misinterpreted at some places
* quoting paths in "src_list.txt" to support paths containing spaces, as some printed paths were absolute and contain those for me.

While some tests till fail, I don't think this has to do with my changes:

	===============================================
	Kaitai Struct Java specs
	Total tests run: 167, Failures: 2, Skips: 0
	===============================================

	#### JavaBuilder: process_status: #<Process::Status: pid 9412 exit 1>
	#### JavaBuilder: running tests: elapsed: 2.14068s

### DIFF
--- a/builder/java_builder.rb
+++ b/builder/java_builder.rb
@@ -8,29 +8,37 @@ class JavaBuilder < PartialBuilder
   def initialize
     super
 
-    @spec_dir = 'spec/java/src'
-    @compiled_dir = "#{@config['FORMATS_COMPILED_DIR']}/java"
-    @project_file = "#{@compiled_dir}/src_list.txt"
+    @spec_dir     = File.join('spec', 'java', 'src')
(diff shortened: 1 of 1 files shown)


GreyCat:
Makes sense, thanks for this effort! I kind of neglected this compatibility issue again when I was rewriting it, thanks for sorting this out!

PR #66, opened 2019-11-12
Migrate to new csproj VS2017+

pluskal:
`AssemblyInfo` is no longer required, info moved to `csproj`. 
`Csproj` do not require listing files so template is also removed.
Adding copy phase do `csharp_builder.rb` so copmpiled sources are copied with CI to test project destination.

Note that project cannot be compiled, same as the master, see [#644](https://github.com/kaitai-io/kaitai_struct/issues/644) for details. After manual resolution of compile errors it seems ok.


builder/csharp_builder.rb
```diff
@@ -53,13 +54,18 @@ def list_disposable_files
   end
 
   def create_project(mand_files, disp_files)
-    tmpl = File.read(@project_template)
-    files_xml = (mand_files + disp_files).map { |x| "    <Compile Include=\"#{x}\" />" }.join("\n")
-    project = tmpl.gsub(/%%%FILES%%%/, files_xml)
-    File.write(@project_file, project)
+    Dir[@compiled_dir + "/**/*.{cs}"].each do |original_parser_filepath| 
+      new_parser_filepath = original_parser_filepath.gsub(@compiled_dir, @spec_dir + "/parsers/")
```
GreyCat: Ok, I see that, given new approach does not list files to compile inside csproj, it means that all files have to be compiled. To work with that paradigm, we will not be just creating project file, but rather maintain a copy of parsers generated in `compiled/csharp`. My questions to you then:

* Is it necessary / mandatory for this copy to be in `@spec_dir`?
* If it is, how do we manage life cycle of that directory?
  * Shall we add it to gitignore to avoid people adding its contents to the repo?
  * How do we ensure that every subsequent attempt of the build will not pick up files copied on previous attempt (which proved to be causing build problems)?
    * Shall we clear this completely first thing in `create_project` + recopy files?
    * Shall we copy all files initially and then delete files proven to be broken after build analysis?

GreyCat:
I also suspect that this won't work with ancient mono's xbuild?

pluskal:
> I also suspect that this won't work with ancient mono's xbuild?

I am not sure about this. I noticed, that `xbuild` is deprecated according to logs in CI.

```
[0K$ ./ci-$TARGET$SUBTARGET
#### CSharpBuilder: initialized
XBuild Engine Version 14.0
Mono, Version 5.10.1.20
Copyright (C) 2005-2013 Various Mono authors
Mono JIT compiler version 5.10.1.20 (tarball Thu Mar 29 10:48:35 UTC 2018)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          supported, not enabled.
	GC:            sgen (concurrent by default)
#### CSharpBuilder: creating project with 339/339 files
#### CSharpBuilder: project file created: "spec/csharp/kaitai_struct_csharp_tests/kaitai_struct_csharp_tests.csproj"
#### CSharpBuilder: build attempt 1 (log: test_out/csharp/build-1.log)
#### CSharpBuilder: running command: ["xbuild", "spec/csharp/kaitai_struct_csharp_tests/kaitai_struct_csharp_tests.csproj"], log: "test_out/csharp/build-1.log"

>>>> xbuild tool is deprecated and will be removed in future updates, use msbuild instead <<<<
```

What about to switch to `msbuild` as the devs suggest?

builder/csharp_builder.rb
pluskal: You are right. I have not realized that you may want to build it multiple times, removing the broken test+parser each iteration.

I will modify this PR.

> Is it necessary / mandatory for this copy to be in @spec_dir?
It can be linked from other directory in the same way how you had it.

pluskal: Now the `csproj` is committed in the repository and by default links all parsers, so it behaves as expected for local development. But, in `create_project` the files are listed explicitly as before.

GreyCat:
There is no msbuild in mono packages, so I would assume that it means it needs to be download and installed separately. I'm 95% sure that it's not just a matter of swapping xbuild => msbuild — likely we'll need to configure msbuild somehow to teach it that we have this installation of mono here and this is what we want to use to compile these .cs files.

builder/csharp_builder.rb
```diff
@@ -53,13 +53,18 @@ def list_disposable_files
   end
 
   def create_project(mand_files, disp_files)
-    tmpl = File.read(@project_template)
-    files_xml = (mand_files + disp_files).map { |x| "    <Compile Include=\"#{x}\" />" }.join("\n")
-    project = tmpl.gsub(/%%%FILES%%%/, files_xml)
+    tmpl = File.read(@project_file)
+    files_xml = (disp_files).map { |x| "    <Compile Include=\"#{x}\" />" }.join("\n")
+    project = tmpl.gsub(/    <Compile Include="..\\..\\..\\compiled\\csharp\\\*\*\\\*.cs" \/>/, files_xml)
     File.write(@project_file, project)
```
GreyCat: What you propose here is generally a very bad thing to do: you programmatically *edit* files checked into the repository. Please keep them separated: either we use checked-in .csproj file without any changes (changing the environment around it, for example), or we clearly show that it's a template and we're generating actual .csproj every time we build it with this builder.

builder/csharp_builder.rb
```diff
@@ -53,13 +53,18 @@ def list_disposable_files
...
     @project_file
   end
 
+  def copy_with_path(src, dst)
+    FileUtils.mkdir_p(File.dirname(dst))
+    FileUtils.cp(src, dst)
+  end
```
GreyCat: If you're not copying anything, this is unused.

spec/csharp/.gitignore
```diff
@@ -253,4 +253,4 @@ paket-files/
 *.sln.iml
 
 # Kaitai Struct test builder specific: this file is generated in `run-csharp`
-kaitai_struct_csharp_tests/kaitai_struct_csharp_tests.csproj
+kaitai_struct_csharp_tests/parsers/*
```
GreyCat: Again, if you're not copying anything, this is not needed.

pluskal:
> There is no msbuild in mono packages, so I would assume that it means it needs to be download and installed separately. I'm 95% sure that it's not just a matter of swapping xbuild => msbuild — likely we'll need to configure msbuild somehow to teach it that we have this installation of mono here and this is what we want to use to compile these .cs files.

I'll try to look at this more closely. It seems that some distributions contain [mono/msbuild](https://github.com/mono/msbuild) package, according to [xamarin.github.io/bugzilla-archives](https://xamarin.github.io/bugzilla-archives/52/52797/bug.html).

builder/csharp_builder.rb
pluskal: I agree that it is bad to modify committed files. But I cannot come up with anything better.

> either we use checked-in .csproj file without any changes (changing the environment around it, for example)

Can you propose how to do it without the template? 

I do not like the template, because using it forces a developer to manually change it and keep track of it not to commit to the repo. Also, a developer cannot open the solution and build it out-of-the-box after cloning it. He needs to figure it out.

I think that it is best to allow a developer to clone and run it out of the box the same way as he is used to on .net platform and does necessary modifications on CI.

GreyCat:
Actually, we've [already broken mono/xbuild builds](https://travis-ci.org/kaitai-io/ci_targets/jobs/609843319) with https://github.com/kaitai-io/kaitai_struct_csharp_runtime/pull/15:

```
Errors:
/home/travis/build/kaitai-io/ci_targets/tests/spec/csharp/kaitai_struct_csharp_tests/kaitai_struct_csharp_tests.csproj (default targets) ->
/usr/lib/mono/xbuild/14.0/bin/Microsoft.Common.targets (ResolveProjectReferences target) ->
	/home/travis/build/kaitai-io/ci_targets/runtime/csharp/kaitai_struct_runtime_csharp.csproj: error : /home/travis/build/kaitai-io/ci_targets/runtime/csharp/kaitai_struct_runtime_csharp.csproj: The default XML namespace of the project must be the MSBuild XML namespace. If the project is authored in the MSBuild 2003 format, please add xmlns="http://schemas.microsoft.com/developer/msbuild/2003" to the <Project> element. If the project has been authored in the old 1.0 or 1.2 format, please convert it to MSBuild 2003 format.
```

builder/csharp_builder.rb
GreyCat: > I do not like the template, because using it forces a developer to manually change it and keep track of it not to commit to the repo.

Why would developer need to manually change a template? The very point of using a template is separation of "template" (which is normally never changed) and "project" (which is changed frequently, being generated in the build process, depending on how well stuff compiles).

> Also, a developer cannot open the solution and build it out-of-the-box after cloning it. He needs to figure it out.

Developer cannot open that solution in any case. At the bare minimum, workflow requires (1) building compiler, (2) building ksy -> csharp files, (3) either cleaning up project manually to exclude things which fail to compile, or invoking partial builder to generate that project for you.

> I think that it is best to allow a developer to clone and run it out of the box the same way as he is used to on .net platform

We support more than a dozen of different languages/platforms. While it's an interesting goal to make KS feel more "natural" for certain target audiences, I don't think we'll realistically be able to satisfy everyone. If we'll aggressively move towards the flow that .NET developers are used to (e.g. integrating compiler / format building into csproj), that might hurt the flow which others are used to.

pluskal: > Why would developer need to manually change a template? The very point of using a template is separation of "template" (which is normally never changed) and "project" (which is changed frequently, being generated in the build process, depending on how well stuff compiles).

>Developer cannot open that solution in any case. At the bare minimum, workflow requires (1) building compiler, (2) building ksy -> csharp files, (3) either cleaning up project manually to exclude things which fail to compile, or invoking partial builder to generate that project for you.

Well, my point was to limit dependencies that a developer needs to install on his machine before he can contribute. The bare minimum required Java, Scala, Ruby, etc. So it is not just a "clone" operation for building and contribution.

> We support more than a dozen of different languages/platforms. While it's an interesting goal to make KS feel more "natural" for certain target audiences, I don't think we'll realistically be able to satisfy everyone. If we'll aggressively move towards the flow that .NET developers are used to (e.g. integrating compiler / format building into csproj), that might hurt the flow which others are used to.

I did not meant to bend KS for .NET developers, but for any developer in any language that KS supports, so a developer would not be required to install dependencies beyond a setup that is common for his language. It might be a troublesome in corporates when you need approvals for SW installation so it might discourage people. 

But I see your point so let me to change it back to the template.

pluskal:
I tried to rebase this PR to the current master and I am not sure about updating the current spec.

What do you think (@GreyCat) about keeping the current **csharp** spec for mono/.net framework and creating a new for .NET core? 

I tried to rebase this PR to current master and test it with msbuild, xbuild and windows with netcli. It requires a lot of if/else path exists checks because of new csproj format and packages.config stores packages in different locations and with different file structure (sln/packages/NUnit.ConsoleRunner.3.4.1/tools/nunit3-console.exe) vs (user profile/.nuget/packages/NUnit.ConsoleRunner/3.4.1). And to make the new csproj work with MSBuild, it requires to run `msbuild -t:restore` before the actual MSBuild run.

With my aim to add async support, https://github.com/kaitai-io/kaitai_struct/issues/640 I would need to increase TargetFramework to netstandard2.0 in [Kaitai Struct Csharp Runtime](https://github.com/kaitai-io/kaitai_struct_csharp_runtime).  Netstandard2.0 is not compatible with netframework4.5 but only with netframework4.6.1+.

I would propose to add a new project to [Kaitai Struct Csharp Runtime](https://github.com/kaitai-io/kaitai_struct_csharp_runtime) containing `async` runtime.

PR #70, opened 2019-12-10
Fix enum generation for values with id bigger than Int.MAX_VALUE

valery1707:
See more at https://github.com/kaitai-io/kaitai_struct_compiler/pull/186


valery1707:
Currently test classes was generated without fixes from https://github.com/kaitai-io/kaitai_struct_compiler/pull/186 - to check is it really needed.

GreyCat:
Thanks, this is very valuable!

PR #73, opened 2020-01-17
Add test str_combine combining different string types

generalmimon:
Demonstrates https://github.com/kaitai-io/kaitai_struct/issues/318

It should pass in all languages (I already fixed it in https://github.com/kaitai-io/kaitai_struct_compiler/commit/cbff3163baf5630d55c07b866097beae0b9edaf3) as it concerns only compiler, but FWIW, it might be best to merge this to ensure that it won't break in the future.

It tests all possible combinations of all string types in KS, so it's quite solid test.


GreyCat:
Makes sense, thanks!

PR #74, opened 2020-01-30
Test calling methods on parentheses with binary operators

generalmimon:
I noticed that the KSC doesn't always wrap expressions in parentheses where neccessary. For instance, `(str_a + str_b).substring(3, 9)` generates the same code as `str_a + str_b.substring(3, 9)` for most languages. It's low priority, but it's clearly a bug in the compiler and should be fixed.

If the bug will take effect or not depends on the actual implementation in the target language, so I think it's wise to add a test for it to see where it works and where not.

kaitai-io/kaitai_struct#69 is related. Simply put, sometimes there are unnecessarily many brackets and sometimes too little. However, unless the compiler is 100% sure that the brackets are useless, it's always better to include them.


GreyCat:
There has been a effort long ago and [relevant branch](https://github.com/kaitai-io/kaitai_struct_compiler/tree/optimal_expressions) to do it in more optimal way. Unfortunately, I wasn't very successful with finishing that work, and never got the chance to return to it.

I think that this one still works good, at least exposing all these problems.

PR #75, opened 2020-02-10
Regen tests from KST for Go + Add Go test panic handling defer func()

generalmimon:
Many tests are missing for Go, so I regenerated all available from KST. I manually reverted all changes that were breaking the tests, and I made sure that I'm not breaking any passing test.

I also noticed that there was some change in https://github.com/kaitai-io/kaitai_struct_tests/commit/27e3785686db0f655287042727f14ff62954a67e#diff-85 that is incorporated in a number of tests, so I also modified the KST translator the same way.


generalmimon:
I've just noticed that there is a explanation for it: doing `./spec_kst_to_all -t all ...` doesn't generate test for Go, because Go is manually excluded from the list of "good languages":

https://github.com/kaitai-io/kaitai_struct_tests/blob/cb4edb86971f714bf56eef66925827015016a77b/translator/src/main/scala/io/kaitai/struct/testtranslator/Main.scala#L32

`git blame` blames commit https://github.com/kaitai-io/kaitai_struct_tests/commit/0590480901f6074350800bd3b87b7420182b6cd7. @GreyCat, do you recall the particular reason for this?

GreyCat:
Invoking GoTranslator when using "go" target for KST was dangerous, as GoTranslator was (and probably is) prone to crashing with exceptions, most notably with NotImplemented errors. These crashes made it hard to do bulk regenerations for all languages.

If that's not the case, we can easily enable it back.

translator/src/main/scala/io/kaitai/struct/testtranslator/specgenerators/GoSG.scala
```diff
@@ -44,6 +45,16 @@ class GoSG(spec: TestSpec, provider: ClassTypeProvider) extends BaseGenerator(sp
...
+    out.puts("defer func() {")
+    out.inc
+    out.puts("if r := recover(); r != nil {")
+    out.inc
+    out.puts("debug.PrintStack()")
+    out.puts("t.Fatal(\"unexpected panic:\", r)")
+    out.dec
+    out.puts("}")
+    out.dec
+    out.puts("}()")
```
GreyCat: This is actually a pretty significant change. Please add it to PR title.

generalmimon: I honestly don't know what is it good for, I just concluded that given that most existing tests are using this and the `./spec_ruby_to_go` is also written in this manner, it's probably wise to do the same in the KST translator.

See commit https://github.com/kaitai-io/kaitai_struct_tests/commit/27e3785686db0f655287042727f14ff62954a67e. I must admit that I don't understand Go much.

GreyCat:
Thanks, this is very big effort! Let's try it at the CI!

PR #76, opened 2020-02-26
Add Lua spec generator to KST translator

generalmimon:
I'm by no means a Lua expert, just followed the approach used in [the existing tests](https://github.com/kaitai-io/kaitai_struct_tests/tree/master/spec/lua).


GreyCat:
That's a massive effort, thanks!

PR #79, opened 2020-03-07
Don't quit whole Lua suite if require error occurs

generalmimon:
It's really stupid to always make sure that in the `run_test_suite.lua` are listed only tests for which the compiler will be able to generate the parser and that the parser won't cause any syntax errors. Otherwise it will break the whole CI build for Lua.

So I added the function `srequire()` that handles the require failure and makes sure that it makes its way as a corresponding test failure.


GreyCat:
Sounds really cool, merging in :)

GreyCat:
One small suggestion: may be we can even escape the need to help the list of files here at all, just globbing everything available?

generalmimon:
@GreyCat
> One small suggestion: may be we can even escape the need to help the list of files here at all, just globbing everything available?

Definitely agree, that's what I was thinking of. This change will help a lot with it, because it's hard to glob only those compiled parsers that contain no syntax errors. And it would be even harder to show the reason of require failure in the CI.

Now it will be basically enough to move all the `srequire()`s into a Git-ignored .lua file (we'll generate it with globbing) and require that script in the `run_test_suite.lua`. And we won't have to worry whether requiring some test triggers error (it'll no longer break the whole build) and we get a clear failure message in our CI what was wrong.

GreyCat:
Why don't we just do a loop with globbing and do `srequire(...)` in that loop right in lua script? I mean, why generate Git-ignored .lua file?

generalmimon:
> Why don't we just do a loop with globbing and do `srequire(...)` in that loop right in lua script?

Oh, I totally missed the possibility to do that directly from the Lua script! Yeah, that'll be a lot easier. When you said globbing, I thought of some Bash script generating `run_test_suite.lua` file, don't know why. That's why I suggested separating the Git-tracked code and untracked list of `srequires` that would be required from the tracked one.

Thanks for showing me the light :grinning:

PR #80, opened 2020-03-11
Integers: test min and max values

cherue:
This changes the integers test to test the min and max values of both signed and unsigned integers.

I made this test while looking into kaitai-io/kaitai_struct_javascript_runtime/pull/11


formats/integers.ksy
```diff
@@ -1,58 +1,44 @@
...
-  - id: sint64be
-    type: s8be
+  - id: unsigned_min
+    type: unsigned
+  - id: unsigned_max
+    type: unsigned
+  - id: signed_min
+    type: signed
+  - id: signed_max
+    type: signed
```
GreyCat: The idea of "integers" test is to test most bare minimum of functionality of KS — that is, bultin integer types. It makes sense to test default endianness here too (and this is why it originally had `endian: le`), but we need to keep it simple, i.e. no subtypes / substructures, no substreams, no variable length structures, etc.

I agree that it might make sense to remove `endian: le` and `contents: ...` from this KSY to make it even simpler, but I'm very much hesitating in adding more stuff (even if that's the stuff we normally take for granted), as it kind of breaks the idea of going from simple to more complicated in development of compilers/translators for new languages.

cherue: I didn't know that was the idea behind the integers test. Makes sense to leave it as is.

How about a separate test like `integers_min_max` instead? Testing these values would properly fail on JS so I think it's a good idea. Might uncover something in other runtimes, too. I'll just go ahead and make a PR.

cherue:
Closed in favor of #81

formats/integers.ksy
GreyCat: Makes sense, thanks again for your contributions and effort!

PR #81, opened 2020-03-12
Add integers_min_max test

cherue:
This tests the min and max values of both signed and unsigned integers.


GreyCat:
Thanks for this effort!

PR #83, opened 2020-03-15
Use assertSame() instead of assertEquals() in PHP tests

generalmimon:
Currently all tests use `assertEquals()` for asserts. However, this method uses equality operator (`==`), which is infamous for its weird type conversions, and it can cause various hidden bugs. For example, `'0012' == '12'`, `"608E-4234" == "272E-3063"`, `3 == true`, `"0" == false`, `"foobar" == 0`, `"true" == 0`,
`"false" == 0` and so on.

Unlike this unpredictable method, `assertSame()` uses operator `===` checking if two values are **identical**, without any type juggling (supposedly has also better performance for this reason), so it's safer to use and it's generally preferred over `assertEquals()` whenever possible. And it did actually catch a type mismatch in the test BitsSimple, there was an assertion that a `seq` field of type `b1` equals `0` (and not `false`), with which the `assertEquals()` had no problem of course.

The only `assertEquals()` usages that are left are those asserting floating point numbers (using absolute delta `1e-6`).


GreyCat:
Very good improvement, thanks!

PR #84, opened 2020-03-16
Improve .gitattributes and .gitignore

dgelessus:
(see commit messages for details)


.gitattributes
```diff
@@ -1,7 +1,16 @@
...
 run-* eol=lf
 spec_* eol=lf
 translator-tests eol=lf
+
+# Explicitly mark binary files for Git
+*.bin binary
+
+# Highlight ksy and kst files as YAML on GitHub
+*.kst linguist-language=yaml
+*.ksy linguist-language=yaml
```
GreyCat: That's actually a good point if we want to start a process registering KSY as legitimate language in GH linguist?

.gitattributes
```diff
@@ -1,7 +1,16 @@
...
+
+# Mark generated files for GitHub
+spec/** linguist-generated
```
GreyCat: I'm somewhat hesitating about this one, but, I guess you're right, ultimately, they all should be generated.

dgelessus: The commits that touch the spec directory make it sound like all of the files are generated from the corresponding KST files. Are some of the files there not generated? If so, it might be a good idea to document that somewhere.

.gitattributes
dgelessus: Good idea! It looks like we meet the requirement of "hundreds of repos" using the language (there are 256 GitHub repos with KSY files in them), so I've opened a PR on the Linguist repo: github/linguist#4830

PR #72, opened 2020-01-05
Add repeat_until_calc_array_type KST spec

generalmimon:
See https://github.com/kaitai-io/kaitai_struct_tests/pull/51, too impatient to wait for @ams-tschoening 🙂

PR #85, opened 2020-04-02
Create nested_type_param.kst

fudgepop01:
See kaitai-io/kaitai_struct#724

I'm not entirely sure how this file works so I tried to craft it the best I could. Judging by the others, it _should_ work, but I'd like some confirmation to be sure


spec/ks/nested_type_param.kst
```diff
@@ -0,0 +1,7 @@
+id: nested_type_param
+data: fixed_struct.bin
+asserts:
+  - actual: seq_block.repeated_thing[0]
+    expected: -52947
+  - actual: seq_block.repeated_thing[1]
+    expected: 1262698832
```
generalmimon, line 7: I don't really like asserting these fields (see https://github.com/kaitai-io/kaitai_struct_tests/commit/eaff6822f59a7be422a2e6fd381a6d579057ae08#commitcomment-38238619), and I think they do not even reflect the intention. The only thing that you want to check here is that the parameter `repeat_count` was successfully passed in the type `test::my_type`, isn't it?

So why you just don't do:

```suggestion
  - actual: seq_block.repeat_count
    expected: 2
```

Beauty in simplicity, isn't it? It's not hiding the intent, and you are just using the bare minimum of features. Because you don't want to check if the first and second item of a dynamically sized-array are equal to something, that's just an indirect consequence of `repeat_count` being greater than or equal to 2.

And for the record, if you will ever need to assert arrays, please check also the length of the array. Because checking just the individual elements would pass even if the `repeat_count` is greater than 2, but that's not what we want.

generalmimon:
And once the KST is good, it's a good habit to run:

```
./spec_kst_to_all -t all nested_type_param -f
./spec_kst_to_all -t go nested_type_param -f
```

You'll immediately know if the .kst is valid or not, and you can review the generated test spec in your favorite language if it looks as you would want. And if not, and you can't get into the desired form by tweaking the KST file, just remove the `// ... generated from KST` comment on the top and modify it to get it work. _But I do not depend on it, if you don't feel comfortable with it (or you just don't want to install `sbt` and JRE), I'll do it myself, cool. Last time I had an nonsense argument about this in https://github.com/kaitai-io/kaitai_struct_tests/pull/72 and I'd rather avoid running into another one 😃_

After running the KST generator, commit the generated test specs, preferably with a commit message `Regen nested_type_param from KST` or something like that and include that in the PR.

spec/ks/nested_type_param.kst
fudgepop01: that makes sense! I wasn't quite aware that you could just test a parameter like that at the time. it's quite a lot better than what I had before. Please let me know your thoughts on the revamped version I just pushed

formats/nested_type_param.ksy
```diff
@@ -1,18 +1,15 @@
...
-      - id: world
-        type: s4
-      - id: repeated_thing
-        type: s4
-        repeat: expr
-        repeat-expr: repeat_count
+      - id: body
+        type: str
+        size: my_len
+        encoding: UTF-8
```
GreyCat, line 15: Please don't omit the trailing newline.

generalmimon, line 15: It's definitely better, but it could be even simpler:

```yaml
meta:
  id: nested_type_param
seq:
  - id: foo
    type: nested_type_param::my_type(5)
types:
  my_type:
    params:
      - id: int_param
        type: u4
    seq: [] # this line can be just omitted
```

My point is, that it's completely fine to leave the `seq` list empty, or leave it out completely, because it's not vital for the test to work.

But I have a unpleasant feeling of being a pedant here, the thing that you've done is fine too, just a little bit more complex than it has to be. If you're satisfied with it, me too. It just seems to me that you don't know it's possible to leave the `seq` empty, but if you feel safer when it actually parses something from the stream, it's OK.

generalmimon:
@fudgepop01 Thanks for this contribution! I've done just some small fixes and generated the test specs to the target languages from KST as I explain above in https://github.com/kaitai-io/kaitai_struct_tests/pull/85#issuecomment-608059310.

PR #86, opened 2020-04-03
Tests for regex support in the valid key

jocelynke:
Only tested on cpp_stl_11, python, java, go, js.


formats/valid_regex.ksy
```diff
@@ -0,0 +1,17 @@
...
+    valid:
+      regex: P[A-Z]{2}K-\d
+  - id: pad1
+    size: 2
+  - id: foo2
+    type: str
+    encoding: ASCII
+    size: 10
+    valid:
+      regex: "[A-Z]{4}p*-+U-+DEF"
```
GreyCat, line 17: This misses end-of-file newline.

spec/ks/valid_regex.kst
```diff
@@ -0,0 +1,2 @@
+id: valid_regex
+data: fixed_struct.bin
```
GreyCat, line 2: Misses newline

GreyCat:
Overall note: we're testing only successful cases here, which makes the test itself effectively a no-op — i.e. just a successful parsing result is ok.

What's much more interesting is to test non-matching cases, i.e. see all `valid_fail_*` tests, which should throw an exception.

jocelynke:
> Overall note: we're testing only successful cases here, which makes the test itself effectively a no-op — i.e. just a successful parsing result is ok.
> 
> What's much more interesting is to test non-matching cases, i.e. see all `valid_fail_*` tests, which should throw an exception.

I added a test for that

PR #89, opened 2020-08-10
lua5.1 compatibility

smarek:


ci-lua
```diff
@@ -1,25 +1,43 @@
...
-mkdir -p "$TEST_OUT_DIR/lua"
+ALLOWED_LUA_VERSIONS=("5.1" "5.2" "5.3")
+CURRENT_LUA_VERSION=""
+
+# try to use $VARIETY from env
+for allowed_version in "${ALLOWED_LUA_VERSIONS[@]}"; do
+  if [[ "${allowed_version}" = "${VARIETY}" ]]; then
+    CURRENT_LUA_VERSION="${VARIETY}"
+  fi
+done
```
GreyCat, line 13: You don't really need bash, arrays and [double brackets](https://www.tldp.org/LDP/abs/html/testconstructs.html#DBLBRACKETS) here. Just a normal sh iteration is totally fine:

```suggestion
ALLOWED_LUA_VERSIONS='5.1 5.2 5.3'
CURRENT_LUA_VERSION=

# try to use $VARIETY from env
for allowed_version in $ALLOWED_LUA_VERSIONS; do
	if [ "${allowed_version}" = "${VARIETY}" ]; then
		CURRENT_LUA_VERSION="${VARIETY}"
	fi
done
```

Also please keep the formatting consistent: this file uses tabs, not spaces.

ci-lua
```diff
@@ -1,25 +1,43 @@
...
+
+if [ -z "${CURRENT_LUA_VERSION}" ]; then
+  # fallback to version 5.1
+  CURRENT_LUA_VERSION="5.1"
+fi
+
+LUA_OUT_DIR="${TEST_OUT_DIR}/lua${CURRENT_LUA_VERSION}"
```
generalmimon, line 20: I don't think that different Lua versions should have separate directories. The current approach is that one language goes only into one directory, the language version doesn't matter. Using a special branch for every target in the [`ci_artifacts` repo](https://github.com/kaitai-io/ci_artifacts/) ensures that there are no conflicts.

See branches [`python/2.7`](https://github.com/kaitai-io/ci_artifacts/tree/python/2.7) and [`python/3.4`](https://github.com/kaitai-io/ci_artifacts/tree/python/3.4) for instance, all test artifacts go to directory `test_out/python` in both cases, ~**not** `test_out/python2`~ or something like that.

ci-lua
```diff
@@ -1,25 +1,43 @@
-#!/bin/sh
+#!/bin/bash
```
generalmimon, line 1:
```suggestion
#!/bin/sh
```

Since you eliminated the Bash brackets in https://github.com/kaitai-io/kaitai_struct_tests/pull/89#discussion_r469587719, I think the shebang can be reverted back to POSIX shell for portability.

ci-lua
smarek: Well, maybe I'm confused by ci-python, ci-python3, ci-python3 scripts, that use different directories. However it seems those are not run in Travis CI, only the ci-python one in differently configured environment.
I've reverted, and hopefuly the https://github.com/kaitai-io/kaitai_ci_ui/pull/8 will suffice to differentiate test targets

kst-adoption-report
```diff
@@ -21,7 +21,7 @@ def glob_spec_files(lang)
   when 'javascript'
     Dir.glob('spec/javascript/test_*.js')
   when 'lua'
-    Dir.glob('spec/lua/test_*.lua')
+    Dir.glob("spec/lua/test_*.lua")
```
generalmimon, line 24:
```suggestion
    Dir.glob('spec/lua/test_*.lua')
```

There's a little reason for this change 🙂

ci-lua
```diff
@@ -2,24 +2,42 @@
...
+for allowed_version in $ALLOWED_LUA_VERSIONS; do
+	if [ "${allowed_version}" = "${VARIETY}" ]; then
+		CURRENT_LUA_VERSION="${VARIETY}"
+	fi
+done
+
+if [ -z "${CURRENT_LUA_VERSION}" ]; then
+  # fallback to version 5.1
+  CURRENT_LUA_VERSION="5.1"
+fi
```
generalmimon, line 18: I'd rather use only `lua` command in this script (~not `lua${CURRENT_LUA_VERSION}`~). The `ci-*` scripts are not exactly designed for selecting the language compiler/interpreter like you did here. This should be done in [`.travis.yml`](https://github.com/kaitai-io/ci_targets/blob/4de55e0c64ea4b1136445e21d7fd6ed053220bc9/.travis.yml#L148-L156) ideally by installing the right `apt` packages or running some commands, eventually in [`prepare-lua`](https://github.com/kaitai-io/ci_targets/blob/4de55e0c64ea4b1136445e21d7fd6ed053220bc9/prepare-lua) script.

If running this ([`prepare-lua:13`](https://github.com/kaitai-io/ci_targets/blob/4de55e0c64ea4b1136445e21d7fd6ed053220bc9/prepare-lua#L13))  will make available only the `lua5.3` command and not `lua` (I don't know, that's a question):

```bash
hererocks lua_install -r^ -l${VARIETY}
```

Can you rather add an alias or symlink the `lua5.3` executable (in `.travis.yml` build matrix or `prepare-lua`) to be available also as `lua`?

spec/lua/run_test_suite.lua
```diff
@@ -4,7 +4,7 @@ local luaunit = require("luaunit")
 local lfs = require("lfs")
 
 function srequire(t)
-    local success, err = xpcall(require, require_err_handler(t), t)
+    local success, err = xpcall(function() require(t) end, require_err_handler(t))
```
generalmimon, line 7: I wonder if this is required to support versions older than Lua 5.3 like Lua 5.1, or it's just for readability?

ci-lua
smarek: Well, the Lua binary selection was already in place, I've only modified it to reflect on Travis env variable VARIETY
see https://github.com/kaitai-io/kaitai_struct_tests/blob/master/ci-lua#L13

If you want to change the way it was, i've just tested and hererocks always installs the lua into `lua_install/bin/lua` regardless of version in my environment. If the corect lua bin detection procedure was always pointless, we can safely get rid of that.

However, it might be useful for development if multiple lua binaries are installed, not through `hererocks`, just sayin

spec/lua/run_test_suite.lua
smarek: This is required for 5.1, xpcall in 5.1 does not take wrapped function arguments separately, and you must wrap it like that.
It's also forward compatible with 5.2 and 5.3 targets

ci-lua
generalmimon: > If you want to change the way it was, i've just tested and hererocks always installs the lua into `lua_install/bin/lua` regardless of version in my environment. If the corect lua bin detection procedure was always pointless, we can safely get rid of that.

Yeah, if that's the case, I'd definitely prefer it.

> However, it might be useful for development if multiple lua binaries are installed, not through `hererocks`, just sayin

Well, I think it's easier to just temporarily symlink the `lua` command than rely on some magic variable, if one needs to run multiple Lua versions locally. Are the Lua versions 5.1, 5.2, and 5.3 really so different that it makes sense to test them all?

smarek: Ok, i'll change the script to only use `lua` binary, no detection whatsoever

They are different, especially 5.1 to 5.3 is big step, and 5.2 is the target we want because of Wireshark, so I'd really like to maintain high compatibility by testing with each version separately

smarek: Also, comparison

**5.3**
```
./run-lua
Ran 210 tests in 0.044 seconds, 195 successes, 9 failures, 6 errors
```

**5.2**
```
./run-lua
Ran 210 tests in 0.038 seconds, 191 successes, 13 failures, 6 errors
```

**5.1**
```
./run-lua
Ran 210 tests in 0.038 seconds, 190 successes, 12 failures, 8 errors
```

ci-lua
```diff
@@ -2,24 +2,17 @@
...
 
-# Detect lua executable
-if lua5.3 -v; then
-	LUA_BIN=lua5.3
-elif lua -v; then
-	LUA_BIN=lua
-else
-	echo "Unable to detect lua executable, bailing out :("
-	exit 1
-fi
```
generalmimon, line 20: I think these two branches should be kept. If there is no `lua` executable, there's no point in doing anything else.

smarek: Almost none of the others ci-* scripts has this, so I don't see the point?

generalmimon:
@smarek For your own convenience, you should never commit directly on `master` branch of your fork. Create a topic branch for every pull request you create. See https://github.com/kaitai-io/kaitai_struct_doc/pull/38#issuecomment-653751666.

smarek:
Thanks, yes i'll do feature branches, once the development here takes off a bit :) I know the drill
Cheers!

PR #90, opened 2020-08-13
fixed reading junit xml reports with invalid UTF-8 sequences

smarek:
currently ci-lua produces invalid bytes under some tests, and reports do not get parsed.
this fixes the issue by replacing the incorrect UTF-8 byte sequences by `?` chars
since those are only in test outputs, it should be acceptable


GreyCat:
Makes sense, thanks!

PR #93, opened 2021-03-08
Test to distinguish equality validation of bytes vs. strings.

ams-tschoening:
This makes sure that a textual error message is thrown instead of HEX-bytes. `valid_fail_eq_str_spec.rb` is created manually to be able to check for concrete error messages, no idea if this is possible using KST already.

https://github.com/kaitai-io/kaitai_struct_ruby_runtime/pull/7


spec/ruby/valid_fail_eq_str_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'valid_fail_eq_str'
+
+RSpec.describe ValidFailEqStr do
+  it 'parses test properly' do
+    expect {
+      r = ValidFailEqStr.from_file('src/fixed_struct.bin')
+    }.to raise_error(
+      # Make extra sure that handling of "byte[]" is not used by looking at the message.
+      Kaitai::Struct::ValidationNotEqualError,
+      a_string_including('"BACK"', '"PACK"')
```
generalmimon, line 10: I would definitely compare the full message. This mere testing whether it contains specific strings constitutes too much creativity, and it's not very well convertible to the KST language-agnostic approach (when we decide to implement it in the KST translator).

BTW, for KSC errors we have [`ErrorMessagesSpec`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/e616927af486416f5494b8ff0f4d4720ed360d39/jvm/src/test/scala/io/kaitai/struct/ErrorMessagesSpec.scala#L18), which tries to compile malformed KSY specs in the [**kaitai_struct_tests** / `formats_err/`](https://github.com/kaitai-io/kaitai_struct_tests/tree/97d68e9847aea0783eee21c27933ebf01fd4e441/formats_err) directory and compares them to the expected error message (for example, [**kaitai_struct_tests** / `formats_err/attr_bad_id.ksy`](https://github.com/kaitai-io/kaitai_struct_tests/blob/97d68e9847aea0783eee21c27933ebf01fd4e441/formats_err/attr_bad_id.ksy#L1)). It also asserts the full message.

Anyway, it's definitely a good idea to extend the tests in the current KS test suite to also ensure that the error messages are consistent between all target languages (because as it will surely turn out, they're not).

generalmimon (changes requested):

generalmimon (approved):
Let's go with this. Thanks for the contribution!

PR #96, opened 2023-03-15
Rust tests

Agile86:
Hello everybody,

my name is [Oleh Dolhov](https://github.com/Agile86), my collegue name is [Vitaly Reshetyuk](https://github.com/revitalyr),

we are from Stroz Friedberg (AON) providing this PR to bring Rust-language support to Kaitai.

Right now we can pass almost all tests, except (4) tests, that's using features like:

1. _parent rewrite_

Rust doesn't support inheritance, so there is no common type.

2. _KStructUnit_, tests like `nav_parent_switch_cast.ksy`, `params_pass_array_struct.ksy`, `params_pass_struct.ksy`.

3. _AnyType_, test like `process_coerce_switch.ksy`.

Please, note, you need to clone [compiler](https://github.com/Agile86/kaitai_struct_compiler/tree/rust_basic_support_v2)  and [runtime/rust](https://github.com/Agile86/kaitai_rust/tree/rust_basic_support_v2) exactly `rust_basic_support_v2` branch.

Best regards,
       Oleh & Vitaly.


generalmimon:
@Agile86 @revitalyr I thank you for your work, but I will now take over and kindly ask you not to proceed with any changes. I have no doubt that you have done a lot of work (i.e. the 'quantity' is large), but across all your changes I come across a lot of nonsensical changes which show that, with all due respect, you do not care much about quality. You often only address the consequences of some problems in a completely wrong (logically flawed) way without looking for the cause of the problem and understanding it.

I will do my best to fix as many sins in your code as possible, but unfortunately I will not be able to fix all of them now. I don't have time to explain everything I don't like about your code, but I don't want you to fix anything anyway (I don't feel there's much chance I'd be happy with your fixes). Just by deciding to eventually accept your code, we are incurring a lot of technical debt going forward, which I personally don't like at all, but there is no other way. Hopefully I'll have time for some major rework in the future.

I would greatly appreciate it if you could take this statement professionally and not take offense. It's not personal, I don't know you, I'm just evaluating the code. But I think you should understand this assessment, because I think you yourself must be aware that your changes are often not well thought out.

generalmimon (commented):
@Agile86 @revitalyr Thank you for this PR as well.

I've had quite a bit of work with this one to be happy with it. To some extent that's fine and understandable, because obviously you can't be as familiar with the project as I am, but I still don't understand why you did certain things. For example:

1. I honestly don't understand the purpose of any of the changes to the KST translator that I reverted in https://github.com/kaitai-io/kaitai_struct_tests/compare/0e6254da0de1065370c6afad410077132abfa900...0ce1363dbf17a36e38ce9db94a3c6996615e3b3b. It looked like you didn't understand how to use the KST translator (in which case you could've asked), or you wanted to use it in a different workflow than what it's designed for, which is fine, but there wasn't any good reason to propose these changes in this PR because it has nothing to do with Rust specifically.

2. I don't understand why you thought that generating `assert_eq!($actStr, 0);` for `expected: 'null'` in a KST spec was OK ([`RustSG.scala:100` at `905f1f2a`](https://github.com/kaitai-io/kaitai_struct_tests/commit/905f1f2aa2065c0f40ec753d46710ba0d684a185#diff-a6bcd12fc56efe3c133df903294fadf9330b65f2374740fa685a8946a45fc0b7R87)):

   ```diff
   @@ -91,11 +84,12 @@ class RustSG(spec: TestSpec, provider: ClassTypeProvider) extends BaseGenerator(
      override def nullAssert(actual: Ast.expr): Unit = {
        val actStr = translateAct(actual)
        finish_panic()
   -    out.puts(s"assertNull($actStr);")
   +    out.puts(s"assert_eq!($actStr, 0);")
        // TODO: Figure out what's meant to happen here
      }
   ```

   Did you see in any other language that we would sweep nullable fields under the rug by just storing `0` in them if they have no value? I don't think so. Even the code we generate for C++ where nullable numeric fields are indistinguishable from non-nullable numeric types (unfortunately, [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional) is only available since C++17, while we target C++98 and C++11) provides [null flags](https://doc.kaitai.io/lang_cpp_stl.html#_null_values) so that it's at least possible to distinguish `0` and `null` (although in a bit ugly way). In Rust, we should of course use `Option<T>` instead.

   I fixed the `nullAssert` method in https://github.com/kaitai-io/kaitai_struct_tests/pull/96/commits/8bdfd0c0547b28f97d730ac6b567664ca3fa6a3e so that it expects the generated code to actually use the `Option<T>` type for nullable fields. Currently it doesn't, so the corresponding tests will fail to build, but that's fine. Tests should accurrately represent the current state of what actually works and what doesn't. They should pass only if the features they test fully work, otherwise they should fail. There's no point in crippling the tests to make them "green" even though the features under test don't work properly - we'd just be lying to ourselves that everything is fine when it's not.

3. Another example of crippling a test just to make it green was in `ExprIntDiv` - this is what it looked like before I finally regenerated it from the KST spec [in `16d44c9c`](https://github.com/kaitai-io/kaitai_struct_tests/pull/96/commits/16d44c9c1177abd72fcfb8ffd9d17e0b86801397#diff-79ba9c9d65e6ed67534fc39c44e343e7634ef461b5b08a1900324e2c27f9b7a3) - https://github.com/kaitai-io/kaitai_struct_tests/blob/16d44c9c1177abd72fcfb8ffd9d17e0b86801397~/spec/rust/tests/test_expr_int_div.rs#L24-L27:

   ```rs
       assert_eq!(*r.div_pos_const().expect("error reading"), 756);
       assert_eq!(*r.div_neg_const().expect("error reading"), -756);
       assert_eq!(*r.div_pos_seq().expect("error reading"), 97130679);
       assert_eq!(*r.div_neg_seq().expect("error reading"), -4072);
   ```

   Again, did you see `-756` and `-4072` in any other language? I don't think so. I would think [`expr_int_div.kst:8-15`](https://github.com/kaitai-io/kaitai_struct_tests/blob/0427d48a0da213241f75f7029cd08217d0e1c7ef/spec/ks/expr_int_div.kst#L8-L15) is clear enough:

   ```yaml
     - actual: div_pos_const
       expected: 756
     - actual: div_neg_const
       expected: -757
     - actual: div_pos_seq
       expected: 97130679
     - actual: div_neg_seq
       expected: -4073
   ```

   I'm really missing what the thought process was ("no one will notice and I'll be able to say that my implementation passes more tests?"), but please don't do this. A failing test is better than faked passing one.

4. Yet another example of crippling a test, this time using `#![allow(overflowing_literals)]` to make the `YamlInts` test pass when it absolutely should not pass and the Rust compiler knows this (https://github.com/kaitai-io/kaitai_struct_tests/pull/96/commits/bf66bbefe71e36903321cb0b0aeeb6853e5dd3f5). This was actually in the compiler as well (https://github.com/kaitai-io/kaitai_struct_compiler/pull/250#discussion_r1741278504) and even the runtime library had something similar (`#![allow(unused)]`, which suppressed a bunch of warnings that had no good reason to be suppressed, see https://github.com/kaitai-io/kaitai_struct_rust_runtime/commit/c7ce8438720e5a5d1f49ee0690a05b6668733463).

   Come on, suppressing lints you don't like is not the solution.

5. I'm generally very skeptical about string manipulation on translated expressions (as I also mentioned in https://github.com/kaitai-io/kaitai_struct_compiler/pull/250#pullrequestreview-2287964384). I'm convinced that we should always treat translated expressions as opaque strings - i.e. never attempt to "see what's inside" in any way. The reason is that string operations can't possibly understand the syntax and semantics of the expression inside the string, so it's highly unreliable, see https://github.com/kaitai-io/kaitai_struct/issues/1125 for instance. Whenever it's needed to "see what's inside the expression", [pattern matching](https://docs.scala-lang.org/tour/pattern-matching.html) with AST objects (or perhaps data types) should be used.

   So I wasn't happy to see the string replacements that used to be in `RustSG.scala` in the KST translator. But surprisingly, I tried removing them and found out that they don't actually do anything (https://github.com/kaitai-io/kaitai_struct_tests/pull/96/commits/b0bd33ed5fabd859d3fd66eb04b527584df606bf#diff-a6bcd12fc56efe3c133df903294fadf9330b65f2374740fa685a8946a45fc0b7). So yeah, I don't understand why they were there, but it doesn't matter now.

6. The method you used to "expect an exception" didn't really make much sense ([`test_eof_exception_bytes.rs:18-22`](https://github.com/kaitai-io/kaitai_struct_tests/blob/da1ffc107f4674c3389002feabe7ea9a35cc5014/spec/rust/tests/test_eof_exception_bytes.rs#L18-L22)):

   ```rs
       if let Err(err) = res {
           println!("expected err: {:?}, exception: EndOfStreamError", err);
       } else {
           panic!("no expected exception: EndOfStreamError");
       }
   ```

   First of all, this doesn't actually test the type of the error as it should (and as it does in all other languages). Second, the `println!` is actually useless, because Rust by default doesn't display stdout of passed tests (see https://blog.ssanj.net/posts/2023-09-24-how-to-see-stdout-when-running-tests-in-rust.html). So the above is equivalent to:

   ```rs
   assert!(res.is_err(), "no expected exception: EndOfStreamError");
   ```

   Which makes it even more visible that it doesn't test or indicate (so that it could be checked manually, which I don't think would be good enough, but would at least make some sense) the error type whatsoever.

   Testing the error type is actually very important - if you don't do that, you're just blindly accepting any error that may arise for any reason (which is very prone to producing false negative test results, because the test might as well just crash with some exception before it even starts, but the test result will be "all good").

   I've seen this before in Lua - some time ago, I decided to replace <code>luaunit.<a href="https://luaunit.readthedocs.io/en/luaunit_v3_2_1/#assertError">assertError</a></code> (which doesn't check the error type) with <code>luaunit.<a href="https://luaunit.readthedocs.io/en/luaunit_v3_2_1/#assertErrorMsgMatches">assertErrorMsgMatches</a></code> in https://github.com/kaitai-io/kaitai_struct_tests/commit/bb20b5abc8023c7e36d608539a8c61a288a34784#diff-335bd0c62964dced90d2d661679316bb2ab4be1e36b80db89f4eaee7ed94bc83. I didn't expect this change to reveal any bug, but it did - the `unable to decide endianness` error could actually never be thrown (such case was treated as big-endian), which I fixed in https://github.com/kaitai-io/kaitai_struct_compiler/commit/f79666f84ad3755fa27182c6958e555e75d75498. And `luaunit.assertError(DefaultEndianExprException.from_file, "src/endian_expr.bin")` didn't fail, because it caught an error that occurred because the arguments to `DefaultEndianExprException.from_file` did not match, so no parsing actually took place - the invocation was supposed to be <code>luaunit.assertError(DefaultEndianExprException.from_file<ins><strong>, DefaultEndianExprException</strong></ins>, "src/endian_expr.bin")</code>.

---

However, I managed to correct (almost) all the wrong things I found, so I'm happy with the result. I think it's quite robust now and there should be few problems with it in the future.

revitalyr:
> I honestly don't understand the purpose of any of the changes to the 
> KST translator that I reverted in 0e6254d...0ce1363 
> <https://github.com/kaitai-io/kaitai_struct_tests/compare/0e6254da0de1065370c6afad410077132abfa900...0ce1363dbf17a36e38ce9db94a3c6996615e3b3b>. 


This doesn't match the help message:

> c.copy(outDir = x)
>
> opt[Unit]('f', "force") action { (x, c) =>
>
> c.copy(outDir = specDir)
>
> } text(s"force overwrite specs in production spec dirs (*_default_*: 
> generate in $defaultOutDir)")
>
>
> opt[String]('f', "force") action { (x, c) =>
>
and what is *_not_* default?

Message ID: 
***@***.***>

PR #97, opened 2023-06-25
First working C Implementation

DarkShadow44:
Needs https://github.com/DarkShadow44/kaitai_struct_c_runtime/tree/mr-c-impl
Needs https://github.com/DarkShadow44/kaitai_struct_compiler/tree/mr-c-impl


jhgorse:
That is what the salt is for... bbl. =)

PR #98, opened 2023-07-22
Python: test on different writable streams

generalmimon:
So far, the generic `test_read_write_roundtrip` test and some specific write tests for EOF errors were only testing on the memory-backed [`BytesIO`](https://docs.python.org/3/library/io.html#io.BytesIO) stream. This is fine in most cases, but it turns out that sometimes it's necessary to test on specific file streams instead, because they have slightly different behavior in certain cases, as evidenced by these issues:

- https://github.com/kaitai-io/kaitai_struct_python_runtime/issues/72
- https://github.com/kaitai-io/kaitai_struct_python_runtime/issues/75

Neither of these issues can be reproduced with `BytesIO`, but they are easy to reproduce with the changes in this PR.

This PR replaces one `test_read_write_roundtrip` test with several tests, each using a different stream. In Python 3, the built-in `open()` and the [`io.open()`](https://docs.python.org/3/library/io.html#io.open) functions are equal (at least in CPython, they even both refer to the same function object, i.e. `(open is io.open) == True`, which I also use in the test code), whereas in Python 2, [`open()`](https://docs.python.org/2/library/functions.html#open) and [`io.open()`](https://docs.python.org/2/library/io.html) are different and they also return streams with slightly different API and behavior of the provided methods.

So these are the new roundtrip test methods in Python 3:

- `test_read_write_roundtrip__InMemoryStream`
- `test_read_write_roundtrip__TemporaryFile_io_open_rdwr`
- `test_read_write_roundtrip__TemporaryFile_io_open_rdwr_nobuf`
- `test_read_write_roundtrip__TemporaryFile_io_open_wronly`
- `test_read_write_roundtrip__TemporaryFile_io_open_wronly_nobuf`

and Python 2 has the same methods plus these:

- `test_read_write_roundtrip__TemporaryFile_builtin_open_rdwr`
- `test_read_write_roundtrip__TemporaryFile_builtin_open_rdwr_nobuf`
- `test_read_write_roundtrip__TemporaryFile_builtin_open_wronly`
- `test_read_write_roundtrip__TemporaryFile_builtin_open_wronly_nobuf`

`InMemoryStream` and `TemporaryFile` refer to the names of helper classes in `spec/python/extra/helpers.py`. `InMemoryStream` is a wrapper for [`io.BytesIO`](https://docs.python.org/3/library/io.html#io.BytesIO), `TemporaryFile` creates (using [`tempfile.mkstemp()`](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp)) and manages a temporary file.

---

There is also a helper class `Pipe` for wrapping a pipe provided by the OS, but it's currently not used. Current implementation of serialization relies on seeking quite a bit - even substreams require seeking support, EOF checks rely on seeking (even though we could skip them in case seeking doesn't work), etc. Therefore, the Python runtime raises `ValueError("writing to non-seekable streams is not supported")` when there's an attempt to write to a non-seekable stream. However, testing on pipes would make sense for parsing, where the only features requiring seeking should be `instances` and `consume: false`. This is left as a future improvement.


generalmimon:
These commits fixed runtime library bugs detected by the newly introduced tests: https://github.com/kaitai-io/kaitai_struct_python_runtime/compare/ebfae5fb55...28de847ca4

generalmimon:
@GreyCat I'm pretty confident that the new tests work as they should (and I've also already reworked the implementation several times to make it as simple as possible, but I'm relatively happy with it now), but I'll leave this PR open for a while in case you want to take a look.

PR #99, opened 2023-07-30
Serialization

GreyCat:


generalmimon:
I'd rather mark this PR as draft to avoid a premature merge until the issue with multiple test methods per test .ksy is resolved: https://github.com/kaitai-io/kaitai_struct/issues/1060#issuecomment-1653660227

It wouldn't be _that_ terrible to have our CI misbehaving for a while, but I prefer to avoid merging unfinished things.

generalmimon (approved):
I think we can merge it now.

PR #105, opened 2024-03-18
C++/STL: recover from aborted tests

generalmimon:
Currently, when a single C++ test aborts at runtime (typically due to a `memory access violation`), the entire test suite is terminated. See [Class `test_observer` > `test_aborted()`](https://www.boost.org/doc/libs/1_84_0/libs/test/doc/html/boost/unit_test/test_observer.html#idm46380-bb):

> Called when a critical error is detected.
>
> The critical errors are mainly the signals sent by the system and caught by the Boost.Test framework. Since the running binary may be in incoherent/instable state, the test execution is aborted and all remaining tests are discarded.

This PR allows the CI to recover from that by detecting the aborted test and running the test executable again excluding the previously aborted tests (using the `--run_test=!<absolute_spec>` option, see [Test unit filtering > Relative specification](https://www.boost.org/doc/libs/1_84_0/libs/test/doc/html/boost_test/runtime_config/test_unit_filtering.html#boost_test.runtime_config.test_unit_filtering.relative_specification)).

### DIFF
--- a/aggregate/boost_test_parser.rb
+++ b/aggregate/boost_test_parser.rb
@@ -3,48 +3,80 @@ require 'rexml/document'
 require_relative 'test_parser'
 
 class BoostTestParser < TestParser
-  def initialize(fn)
-    @doc = REXML::Document.new(File.read(fn))
+  def initialize(fn, glob = false)
+    @docs =
+      if glob
+        Dir.glob(fn).map { |x|
+          REXML::Document.new(File.read(x))
+        }
+      else
+        [REXML::Document.new(File.read(fn))]
+      end
   end
 
   def each_test
-    @doc.root.elements.each('TestSuite') { |ts|
-      ts.elements.each('TestCase') { |tc|
-        name = tc.attribute('name').value
+    @docs.each do |doc|
+      r = doc.root
+      next unless r
 
-        raise "Unable to parse name: \"#{name}\"" unless name =~ /^test_(.*?)$/
-        name = underscore_to_ucamelcase($1)
+      r.elements.each('TestSuite') { |ts|
+        ts.elements.each('TestCase') { |tc|
+          name = tc.attribute('name').value
 
-        failures = []
-        tc.elements.each('Error') { |err|
-          failures << TestResult::Failure.new(
-            err.attribute('file'),
-            err.attribute('line'),
-            err.text,
-            nil
-          )
+          raise "Unable to parse name: \"#{name}\"" unless name =~ /^test_(.*?)$/
+          name = underscore_to_ucamelcase($1)
+
+          failures = []
+          tc.elements.each('Error') { |err|
+            failures << TestResult::Failure.new(
+              err.attribute('file'),
+              err.attribute('line'),
+              err.text,
+              nil
+            )
+          }
+          tc.elements.each('Exception') { |err|
+            failures << TestResult::Failure.new(
+              err.attribute('file'),
+              err.attribute('line'),
+              err.text,
+              nil
+            )
+          }
+
+          if failures.empty?
+            status = :passed
+            failure = nil
+          else
+            status = :failed
+            failure = failures[0]
+          end
+
+          # `<TestCase>` elements of tests that are disabled (using the
+          # `--run_test=!<test_name>` command-line option) because they aborted
+          # (typically due to a memory access violation, which terminates the
+          # entire test suite) in a previous run attempt look like this in the
+          # XML log:
+          #
+          # ```xml
+          #     <TestCase name="test_debug_array_user" skipped="yes" reason="disabled"/>
+          # ```
+          #
+          # As you can see, they don't contain any `<TestingTime>` tag, so we
+          # need to make sure we handle this case gracefully.
+          testing_time_node = tc.elements['TestingTime']
+          testing_time =
+            if testing_time_node.nil?
+              0.0
+            else
+              testing_time_node.text.to_f
+            end
+
+          tr = TestResult.new(name, status, testing_time, failure)
+          yield tr
         }
-        tc.elements.each('Exception') { |err|
-          failures << TestResult::Failure.new(
-            err.attribute('file'),
-            err.attribute('line'),
-            err.text,
-            nil
-          )
-        }
-
-        if failures.empty?
-          status = :passed
-          failure = nil
-        else
-          status = :failed
-          failure = failures[0]
-        end
-
-        tr = TestResult.new(name, status, tc.elements['TestingTime'].text.to_f, failure)
-        yield tr
       }
-    }
+    end
   end
 
   def aborted_tests
@@ -59,29 +91,33 @@ class BoostTestParser < TestParser
   end
 
   def each_aborted_test
-    r = @doc.root
-    r.elements.each('TestSuite/TestCase') { |tc|
-      next if tc.elements['Exception'].nil?
+    @docs.each do |doc|
+      r = doc.root
+      next unless r
 
-      aborted = false
+      r.elements.each('TestSuite/TestCase') { |tc|
+        next if tc.elements['Exception'].nil?
 
-      tc.elements.each('Message') { |msg|
-        next unless msg.text == 'Test is aborted'
+        aborted = false
 
-        file_attr = msg.attribute('file')
-        next if file_attr.nil?
+        tc.elements.each('Message') { |msg|
+          next unless msg.text == 'Test is aborted'
 
-        file = file_attr.value
-        next unless file.end_with?('/boost/test/impl/unit_test_log.ipp')
+          file_attr = msg.attribute('file')
+          next if file_attr.nil?
 
-        aborted = true
-        break
+          file = file_attr.value
+          next unless file.end_with?('/boost/test/impl/unit_test_log.ipp')
+
+          aborted = true
+          break
+        }
+
+        next unless aborted
+
+        name = tc.attribute('name').value
+        yield name
       }
-
-      next unless aborted
-
-      name = tc.attribute('name').value
-      yield name
-    }
+    end
   end
 end
--- a/aggregate/boost_test_parser.rb
+++ b/aggregate/boost_test_parser.rb
@@ -46,4 +46,42 @@ class BoostTestParser < TestParser
       }
     }
   end
+
+  def aborted_tests
+    tests = []
+    # An "aborted" test means that "a critical error is detected", which means
+    # that "the test execution is aborted and all remaining tests are discarded"
+    # (see https://www.boost.org/doc/libs/1_84_0/libs/test/doc/html/boost/unit_test/test_observer.html#idm46380-bb).
+    # So in practice there will probably never be more than 1 aborted test (per
+    # log file), but we don't rely on that here and return an array anyway.
+    each_aborted_test { |t| tests << t }
+    tests
+  end
+
+  def each_aborted_test
+    r = @doc.root
+    r.elements.each('TestSuite/TestCase') { |tc|
+      next if tc.elements['Exception'].nil?
+
+      aborted = false
+
+      tc.elements.each('Message') { |msg|
+        next unless msg.text == 'Test is aborted'
+
+        file_attr = msg.attribute('file')
+        next if file_attr.nil?
+
+        file = file_attr.value
+        next unless file.end_with?('/boost/test/impl/unit_test_log.ipp')
+
+        aborted = true
+        break
+      }
+
+      next unless aborted
+
+      name = tc.attribute('name').value
+      yield name
+    }
+  end
 end
--- a/builder/cpp_builder.rb
+++ b/builder/cpp_builder.rb
@@ -3,6 +3,7 @@ require 'set'
 
 require_relative 'partial_builder'
 require_relative 'shellconfig'
+require_relative '../aggregate/boost_test_parser'
 
 class CppBuilder < PartialBuilder
   attr_accessor :mode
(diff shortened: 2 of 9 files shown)


generalmimon:
The recovery works as expected on target `cpp_stl_11/gcc11-linux-x86_64` (`./docker-ci -t cpp_stl -u _11 -i gcc11`). However, it does not work on `cpp_stl_11/clang3.4-linux-x86_64` (`./docker-ci -t cpp_stl -u _11 -i clang3.4`):

```
#### CppBuilder: running tests
#### CppBuilder: run attempt 1 (log: /tests/test_out/cpp_stl_11/results-1.xml)
#### CppBuilder: running command: ["/tests/compiled/cpp_stl_11/bin/ks_tests",
  "--log_format=XML", "--log_sink=/tests/test_out/cpp_stl_11/results-1.xml",
  "--log_level=all", "--report_level=detailed"],
  log: "/tests/test_out/cpp_stl_11/test_run-1.stdout"

Test suite "C++/STL test units for Kaitai Struct" aborted with:
  160 assertions out of 161 passed
  1 assertion out of 161 failed
  32 test cases out of 226 passed
  1 test case out of 226 failed
  193 test cases out of 226 skipped
  1 test case out of 226 aborted

  ...

  Test case "test_instance_std" passed with:
    1 assertion out of 1 passed

  Test case "test_debug_array_user" aborted with:
    1 assertion out of 1 failed

  Test case "test_imports_abs_abs" skipped due to test aborting
  ...
  Test case "test_instance_std_array" skipped due to test aborting
#### CppBuilder: process_status: #<Process::Status: pid 2095 exit 201>
#### CppBuilder: run attempt 1 aborted (1 tests aborted)
#### CppBuilder: run attempt 2 (log: /tests/test_out/cpp_stl_11/results-2.xml)
#### CppBuilder: running command: ["/tests/compiled/cpp_stl_11/bin/ks_tests",
  "--log_format=XML", "--log_sink=/tests/test_out/cpp_stl_11/results-2.xml",
  "--log_level=all", "--report_level=detailed", "--run_test=!test_debug_array_user"],
  log: "/tests/test_out/cpp_stl_11/test_run-2.stdout"
Test setup error: no test cases matching filter
#### CppBuilder: process_status: #<Process::Status: pid 2098 exit 200>
```

As you can see, as soon as the `--run_test=!test_debug_array_user` option is added, the Boost.Test executable fails with:

```
Test setup error: no test cases matching filter
```

It seems that this is because the `{clang3.4,gcc4.8}-linux-x86_64` targets use Boost 1.54.0 (as I mention [here](https://github.com/kaitai-io/kaitai_struct_tests/commit/d07deb4cbbaea4c9e70fe4dd59de30fd09f31733)), which is a version [from 2013](https://www.boost.org/users/history/version_1_54_0.html) and apparently doesn't support test exclusion yet (according to the [docs for Boost Test 1.54.0](https://www.boost.org/doc/libs/1_54_0/libs/test/doc/html/utf/user-guide/runtime-config/run-by-name.html), it supports the `--run_test` option, but only with positive specifications).

I see two ways to solve this problem:

1. Don't use `--run_test=!<test_name>` to exclude tests, instead use another approach compatible with Boost 1.54.0.
2. Install some newer version of Boost in [`clang3.4`](https://github.com/kaitai-io/kaitai_struct_docker_images/blob/ef0ad6e321d86839c7a929a5d5221bfcf1099531/src/cpp_stl/clang3.4-linux-x86_64/Dockerfile) and [`gcc4.8`](https://github.com/kaitai-io/kaitai_struct_docker_images/blob/ef0ad6e321d86839c7a929a5d5221bfcf1099531/src/cpp_stl/gcc4.8-linux-x86_64/Dockerfile) that supports `--run_test=!<test_name>`.

I'd prefer option 2, but of course that will be a little more complicated than just [`apt-get install libboost-test-dev`](https://github.com/kaitai-io/kaitai_struct_docker_images/blob/ef0ad6e321d86839c7a929a5d5221bfcf1099531/src/cpp_stl/gcc4.8-linux-x86_64/Dockerfile#L14). We'll probably have to build Boost.Test from source (as we already do [with CMake](https://github.com/kaitai-io/kaitai_struct_docker_images/blob/ef0ad6e321d86839c7a929a5d5221bfcf1099531/src/cpp_stl/gcc4.8-linux-x86_64/Dockerfile#L21-L32)). A good [tip from SO](https://stackoverflow.com/a/45276859/12940655) is to build only the libraries we need (in our case only the Boost.Test library, probably only for static linking) instead of building everything in the Boost suite, which might take a long time.

aggregate/boost_test_parser.rb
```diff
@@ -3,47 +3,121 @@
...
+          #
+          # As you can see, they don't contain any `<TestingTime>` tag, so we
+          # need to make sure we handle this case gracefully.
+          testing_time_node = tc.elements['TestingTime']
+          testing_time =
+            if testing_time_node.nil?
+              0.0
+            else
+              testing_time_node.text.to_f
+            end
```
GreyCat, line 73: Technically, if we don't mind having nulls in `testing_time`, this will be a tiny bit more concise way to do it:

```suggestion
testing_time = tc.elements['TestingTime']&.text&.to_f
```

or, if we absolutely need it to be `0.0`, then

```suggestion
testing_time = tc.elements['TestingTime']&.text&.to_f || 0.0
```

Not sure about readability of it, though.

generalmimon: > Not sure about readability of it, though.

I know about this operator, but yeah, I don't think it's very readable. Maybe

```rb
tc.elements['TestingTime'] && tc.elements['TestingTime'].text.to_f
```

... would be better.

Either way, I don't see much value in this, given that it works the same as the original version which is well readable though somewhat verbose. But if you want, feel free to refactor.

Just noting that the `&.` operator was introduced in Ruby 2.3, so I'm not 100% sure that using it will not break the `ruby/1.9-linux-x86_64` target that we still have (although it's a `ruby` target and not `cpp_stl`, it will still run `aggregate/convert_to_json` like any other language, which [requires `boost_test_parser`](https://github.com/kaitai-io/kaitai_struct_tests/blob/ac7dd7f068a5ea8e3fb86b2ec9666ff12aeb5b76/aggregate/convert_to_json#L12) and I don't know whether merely the presence of `&.` in the source code wouldn't be a syntax error in Ruby 1.9).

builder/cpp_builder.rb
```diff
@@ -304,33 +305,58 @@ def file_to_test(filename)
   end
 
   def run_tests
-    xml_log = "#{@abs_cpp_test_out_dir}/results.xml"
+    attempt = 1
```
GreyCat, line 308: This solution adds yet another trying loop with its own attempts counter, etc.

As alternative idea: will it bring any benefits we detect offending (aborted) test and then feed that information back to normal partial builder mechanism to exclude offending files out of the binary to get us the stable set of tests in one binary in the end?

On one hand:

* this avoids looking into boost::test invocation flags which might be not compatible with boost versions you've mentioned
* this requires test parsing to determine offending test, but you did that already

On the other hand, I see that such a feedback mechanism from `run_tests` does not exist now and inventing it on whole `partial_builder.rb` might be trickier...

builder/cpp_builder.rb
```diff
@@ -304,33 +305,58 @@ def file_to_test(filename)
...
-      raise "Unknown mode=#{@mode}"
-    end
+      tests_cli = [
+        ks_tests_bin,
+        '--log_format=XML',
+        "--log_sink=#{xml_log}",
+        '--log_level=all',
+        '--report_level=detailed',
+      ]
+      excluded_tests.each { |test| tests_cli << "--run_test=!#{test}" }
```
GreyCat, line 336: This one is also a bit dangerous on Windows: I recall that command line was limited to 32KiB. If we'll concatenate all test names we have now ~7KiB worth of such strings, but it potentially plants an unpleasant surprise for the future.

The only way to work that around as far as I remember is for executable to provide alternative means to source command line args (e.g. read them from file), but boost::test does not seem to support that.

builder/spec/cpp/cpp_builder_spec.rb
```diff
@@ -1,4 +1,5 @@
 require_relative '../../cpp_builder'
+require_relative '../../../aggregate/boost_test_parser'
```
GreyCat, line 2: Nit: I believe this one is not really necessary?

valgrind-cpp_stl
```diff
@@ -12,4 +12,36 @@ CPP_TEST_OUT_DIR="$2"
...
+	run_valgrind "${run_attempt}" "$@" || :
+
+	# if the `aborted_tests-{N}.txt` file is empty, then there were no aborted
+	# tests and it's the last attempt we need to process
+	read_next=
+	while IFS= read -r line; do
+		read_next=1
+		set -- "$@" "--run_test=!${line}"
+	done < "${CPP_TEST_OUT_DIR}/aborted_tests-${run_attempt}.txt"
+done
```
GreyCat, line 43: Again, one more argument towards re-doing the builds excluding offending file after run: we won't need  to maintain such complex magic around valgrind runs.

builder/cpp_builder.rb
generalmimon: In theory, this is a valid concern (I thought of that too, it just didn't feel that important to me) and not only on Windows, Unix-like OSes have a limit too, though probably higher than 32 KiB. In practice, I believe we are relatively far from reaching that limit (as you mentioned, even if every single test aborted, which already sounds extraordinary, we would still be well below that limit).

As you noted, I also haven't found a way to make Boost.Test read the arguments from file. But I see two ways we could solve this problem:

1. Even if Boost.Test itself doesn't support reading CLI args from file, we could probably implement this functionality ourselves by [customizing the module's entry point](https://www.boost.org/doc/libs/1_84_0/libs/test/doc/html/boost_test/adv_scenarios/shared_lib_customizations/entry_point.html). The new entry point would simply read the arguments from a file and pass them to the actual `main` function provided by Boost.Test.
2. We could take inspiration from how `xargs` works, i.e. splitting the long list of arguments into multiple batches so each batch is below the maximum limit. We cannot reliably do this with the exclusion list, but we can with an inclusion list - so we would use the [absolute specification](https://www.boost.org/doc/libs/1_84_0/libs/test/doc/html/boost_test/runtime_config/test_unit_filtering.html#boost_test.runtime_config.test_unit_filtering.absolute_specification), positively listing each test to run.

   This approach has a number of downsides compared to approach 1 - the `cpp_builder` would have to keep track of the entire set of test names, find out the command line length limit (or use `xargs` to take care of that, although that introduces a dependency) and the number of tests that we can run at once (i.e. in one batch) would be limited by how many we can list in the command line, so the test executable would have to be run multiple times just to execute a single snapshot of the test set.

Approach 1 seems much better and it should be pretty easy to implement, probably just a few lines of code.

But I don't see it as a high priority right now because it's not currently causing a problem. My view of our testing system is that it must be only as good as it proves necessary, and no better, because it's an internal tool, not meant be used by end users. We don't need to support C++ compilers / Boost versions that we don't run, or situations that have never happened or cannot happen with the current state of the project.

You know that I'm a perfectionist and I'd rather see everything working in the general case, but when it comes to the testing system, even I realize that solving all potential problems "in advance" doesn't make any difference (it's better to focus our efforts to pretty much any other KS component), only solving real problems does.

builder/cpp_builder.rb
generalmimon: > This solution adds yet another trying loop with its own attempts counter, etc.

Yeah. I don't think you can eliminate it completely in principle, because you don't know in advance which tests will abort, and as soon one test aborts, no more tests are run (this is by design, Boost has no "continue running the suite despite an aborted test" option, that's not a thing). If you want them to run, you have no option but to run the test executable again without that offending test.

> As alternative idea: will it bring any benefits we detect offending (aborted) test and then feed that information back to normal partial builder mechanism to exclude offending files out of the binary to get us the stable set of tests in one binary in the end?

I briefly considered this approach as well, but don't see any benefits. In my view, it would only be more complicated and less flexible. As I explain above, you have to run the test suite `N + 1` times in any case (where `N` is the number of aborted tests), only now with the unnecessary extra step of repeating the linking the object files into the `ks_tests` executable.

> * this avoids looking into boost::test invocation flags which might be not compatible with boost versions you've mentioned

This is no longer a problem because since https://github.com/kaitai-io/kaitai_struct_docker_images/commit/1ca846db680421fde3f6ad5bdff62b61bac5ed11 the `{clang3.4,gcc4.8}-linux-x86_64` targets build the latest Boost 1.84 from source (instead of using ancient the Boost 1.54), which works like a charm. None of Boost versions we use at the moment have a problem with `--run_test=!<test_name>` anymore (as evidenced by https://ci.kaitai.io/ - coincidentally, all `cpp_stl_{98,11}` targets currently have at least one aborted test `NavParentRecursive`, yet they all recovered successfully), so there's no need to avoid it.

---

An actual disadvantage of the approach you're suggesting is that with the "stable" test binary at the end, you won't be able to run the aborted tests via Valgrind after `CppBuilder` has finished. This means that the only information about an aborted will be the message from when the OS signal was caught by Boost, i.e.:

```
memory access violation at address: 0x00000008: no mapping at fault address
```

... which isn't really helpful to diagnose the error - there's no file name, no line number, no stack trace. Valgrind can give you all that, if you only let it run the aborted test (`cpp_stl_11/gcc11-linux-x86_64` > [test_out/cpp_stl_11/valgrind-1.xml:36-57](https://github.com/kaitai-io/ci_artifacts/blob/4ee6502cedb050b22cd0c9cde009006947c53fa8/test_out/cpp_stl_11/valgrind-1.xml#L36-L57)):

```xml
<error>
  <unique>0x0</unique>
  <tid>1</tid>
  <kind>InvalidRead</kind>
  <what>Invalid read of size 8</what>
  <stack>
    <frame>
      <ip>0x54235A</ip>
      <obj>/tests/compiled/cpp_stl_11/bin/ks_tests</obj>
      <fn>debug_array_user_t::cat_t::_read()</fn>
      <dir>/tests/compiled/cpp_stl_11</dir>
      <file>debug_array_user.cpp</file>
      <line>37</line>
    </frame>
    <frame>
      <ip>0x542205</ip>
      <obj>/tests/compiled/cpp_stl_11/bin/ks_tests</obj>
      <fn>debug_array_user_t::_read()</fn>
      <dir>/tests/compiled/cpp_stl_11</dir>
      <file>debug_array_user.cpp</file>
      <line>20</line>
    </frame>
```

(Valgrind logs of aborted tests are currently produced, but not processed by `aggregate/convert_to_json` - I left this as a future improvement.)

valgrind-cpp_stl
generalmimon: We don't _need_ to do that now either - as far as the contents of `ci.json` is concerned at the moment, only the log from the last Valgrind run (excluding any aborted tests) is copied to `valgrind.xml` and [used by `aggregate/convert_to_json`](https://github.com/kaitai-io/kaitai_struct_tests/blob/ac7dd7f068a5ea8e3fb86b2ec9666ff12aeb5b76/aggregate/convert_to_json#L71). This is explained at the bottom of `valgrind-cpp_stl` ([`valgrind-cpp_stl:45-47`](https://github.com/kaitai-io/kaitai_struct_tests/blob/ac7dd7f068a5ea8e3fb86b2ec9666ff12aeb5b76/valgrind-cpp_stl#L45-L47)):

```sh
# Currently, aggregate/convert_to_json only processes "valgrind.xml", so we
# make a copy with that name of the last (successful) attempt's log
cp -v "$CPP_TEST_OUT_DIR/valgrind-${run_attempt}.xml" "$CPP_TEST_OUT_DIR/valgrind.xml"
```

However, we'll eventually want to aggregate all `valgrind-*.xml` logs that this new version of `valgrind-cpp_stl` produces to `ci.json`, because these Valgrind logs from running aborted tests contain all the details of the memory error causing the abortion, as I mentioned above in https://github.com/kaitai-io/kaitai_struct_tests/pull/105#discussion_r1536646230.

So although the generated `valgrind-*.xml` logs are not used at the moment, they can be already helpful when manually inspecting the CI artifacts, and it doesn't hurt to have them available.

builder/spec/cpp/cpp_builder_spec.rb
generalmimon: Well, I added some tests to check that the detection of aborted tests works:

https://github.com/kaitai-io/kaitai_struct_tests/blob/25b2182b24ba7e1bbc5514e9a676120897930ee0/builder/spec/cpp/cpp_builder_spec.rb#L438-L447

I realize that it's not very clean to add a test of [`aggregate/boost_test_parser.rb`](https://github.com/kaitai-io/kaitai_struct_tests/blob/ac7dd7f068a5ea8e3fb86b2ec9666ff12aeb5b76/aggregate/boost_test_parser.rb) to [`builder/spec/cpp/cpp_builder_spec.rb`](https://github.com/kaitai-io/kaitai_struct_tests/blob/ac7dd7f068a5ea8e3fb86b2ec9666ff12aeb5b76/builder/spec/cpp/cpp_builder_spec.rb), but I was a bit lazy to create a new `*_spec.rb` file under the `aggregate/` folder, sorry :) Feel free to move these `BoostTestParser` tests to the right place.

GreyCat: Got it, makes sense then!

aggregate/boost_test_parser.rb
GreyCat: Yep, agree. Forgot about that Ruby 1.9 aspect completely.

PR #111, opened 2024-04-06
C#: slightly tweak KST generator to generate more automatic tests

Mingun:
This PR generates casts of numerical expected values to enumerations if the actual value of the enum type.

This PR depends on https://github.com/kaitai-io/kaitai_struct_compiler/pull/296 and should be merged only after it.


translator/src/main/scala/io/kaitai/struct/testtranslator/specgenerators/CSharpSG.scala
```diff
@@ -57,14 +58,24 @@ class CSharpSG(spec: TestSpec, provider: ClassTypeProvider) extends BaseGenerato
...
+    // If type of value is enum but we check it against numeric value,
+    // do the cast of expected value to enum
+    expStr = (actType, expType) match {
+      case (act: EnumType, _: IntType) => {
+        val casted = CSharpCompiler.kaitaiType2NativeType(act, true)
+        s"($casted) $expStr"
+      }
+      case _ => expStr
     }
+    // TODO: fix order - actually it is (expected, actual)
```
Mingun, line 77: It also possible to address two TODOs in this PR about order of parameters in `Assert.AreEquals`.

One way is to change order in the generator, but the generated files would look awkward (the "actual, then expected" order is better at my sight). The other way is to implement our own `AreEqual` methods in the `CommonSpec` with swapped order of arguments and generate code with them, but I'm afraid the failed assertion then would link to the helper method instead of actual place of failure.

What would you prefer?

generalmimon: @Mingun:

> One way is to change order in the generator, but the generated files would look awkward

Let's go with this. In the end, it doesn't really matter what the test specs generated by the KST translator look like. They just have to be functional. It seems to be much simpler to follow the "when in Rome, do as the Romans do" approach (the only issue is that we have to migrate all existing C# specs that now use the wrong order, but that's just one-time fix and then it'll be correct forever), so I don't see any reason not to.

The "reverse the order at runtime" approach you came up with:

https://github.com/kaitai-io/kaitai_struct_tests/blob/d588602aecdd49cdae08f67a80e78e803e5c4d3f/spec/csharp/kaitai_struct_csharp_tests/CommonSpec.cs#L91-L136

... works too, but it seems unnecessary to me. And we'll probably have to keep adding more methods/overloads in the future just to stubbornly maintain the "actual, expected" order in the (mostly) auto-generated test specs that very few people actually look at.

> (the "actual, then expected" order is better at my sight).

This is really just a matter of habit. There is no right or wrong order, just what some are more used to and the other.

I know that Rust for example deliberately doesn't call `assert_eq!` parameters "expected" and "actual", but just "left" and "right". Some Rust programmers use it as "actual, expected" and others as "expected, actual", and it doesn't matter.

However, I think we indeed get it right for those test frameworks (like NUnit in C#) that literally say which value is "expected" and which is "actual" when the assertion fails, so the error details at https://ci.kaitai.io/ are on point and not confusing (as they are right now).

Mingun: I changed the order of arguments and regenerated C# tests.

Mingun:
@generalmimon, is there something wrong with this PR? If yes, could you tell me what?

PR #112, opened 2024-04-06
Rename `spec_kst_to_all` to `build-tests` and better document process of testing

Mingun:
In the https://github.com/kaitai-io/kaitai_struct_tests/pull/108#issuecomment-2028473115 @generalmimon was not against to removing existing `build-tests` script which is not very useful. On the other hand, `spec_kst_to_all` script which is actually useful:
- named unclear
- contains underscores which are slightly inconvenient to enter (requires pressing SHIFT + `-` in QWERTY layout)

This PR solves both problems by removing the old `build-tests` script and renaming `spec_kst_to_all` script to `build-tests`. It also updates documentation, because I find it is difficult to have documentation of how to run tests far away from tests itself in other repository (https://github.com/kaitai-io/kaitai_struct_doc/blob/c87ae4cb447abfb043bcbade1c8d601037469142/kst.adoc).

I also described directories that was not described to have the full picture in the Readme.

This PR should be merged together with https://github.com/kaitai-io/kaitai_struct_doc/pull/55 to update mentioned `kst.adoc`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,29 @@
 # Kaitai Struct: specs and tests
 
-This repository contains specifications and tests for 
+This repository contains specifications and tests for
 [Kaitai Struct](https://github.com/kaitai-io/kaitai_struct) project.
 
 ## What's inside
 
 The repository is laid out like that:
 
-* `src/` - binary input files that would be parsed during the tests
+* `aggregate` - scripts that converts result of tests into unified `ci.json` output,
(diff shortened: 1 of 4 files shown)


Mingun:
@generalmimon , is it possible to merge this PR? It greatly simplifies developing process, because you should not remember to make `sbt publishLocal` after building compiler.

PR #114, opened 2024-04-07
Implement testing of `to-string` implementation via KST

Mingun:
- Added a new key to KST -- `to-string` which should contain the string which is expected from `to-string` implementation.
- Added missed tests for Nim and Rust

Depends on https://github.com/kaitai-io/kaitai_struct_compiler/pull/297 which introduces `CppTranslator.doRawStringLiteral` method to get a plain old C string literal instead of `std::string` object.


generalmimon:
I don't see any benefit in implementing this on the side of the KST translator when we'll eventually need this in KSC anyway - see https://github.com/kaitai-io/kaitai_struct/issues/980 (`-webide-representation` already has this feature, so if we want `to-string` to be a visualizer-independent full-fledged replacement for `-webide-representation`, we're going to need it).

So instead of introducing this unnecessary `to-string` key to KST specs, it would be better to implement `.to_s` for structs with `to-string` on the KSC's side, and then we'll just do `actual: struct.to_s` or something.

Needless to say that your proposed `to-string` key only works on the top-level object (there is no way to specify which object we want to convert to string using `actual: ...`), which is an unnecessary limitation.

Mingun:
@generalmimon, I'm not sure, how it would be possible to implement in non-ambiguous way.
```yaml
id: to_string_custom
data: term_strz.bin
asserts:
  - actual: to_s
    expected: '"s1 = foo, s2 = bar"'
```
would mean that object should have `to_s` seq field or instance. What the syntax you'll propose for calling `to-string`?

PR #115, opened 2024-04-08
Python: and compatibility layer for Python 2 & <=3.11

Mingun:
Use the non-deprecated method in Python 3 and fix tests in Python >=3.12:
- Python 2 has only `assertRaisesRegexp`
- Python <=3.11 have both `assertRaisesRegex` and (deprecated) `assertRaisesRegexp`
- Python >=3.12 has only `assertRaisesRegex`

Should fix the following tests on https://ci.kaitai.io/

- EofExceptionU4
- EosExceptionBytes
- EosExceptionU4
- ExprBytesNonLiteral

Closes https://github.com/kaitai-io/kaitai_struct/issues/1104


spec/python/test_eof_exception_u4.py
```diff
@@ -3,8 +3,15 @@
 
 from eof_exception_u4 import EofExceptionU4
 
+# Python 2 has only assertRaisesRegexp method
+# Python 3.11 and below have both assertRaisesRegex and assertRaisesRegexp
+# Python 2.12 and above has only assertRaisesRegex
+# The tests run on both Python 2 & 3, so we add missing method for compatibility
+if not hasattr(unittest.TestCase, 'assertRaisesRegex'):
+    setattr(unittest.TestCase, 'assertRaisesRegex', unittest.TestCase.assertRaisesRegexp)
```
GreyCat, line 11: Sounds good, appreciate your fix here!

Do you think we can actually extract this into a separate file not to repeat these lines everywhere?

Something akin to [helpers/javascript](https://github.com/kaitai-io/kaitai_struct_tests/tree/master/helpers/javascript)?

Mingun: I was thinking about building a base class for tests, but since I don't test it (I'm too lazy to set up all the language environments that will only be needed for one thing), I was not sure that this will work and I made the simplest solution.

Something like this, but I'm not sure if constructor should have some arguments:
```python
class BaseTest(unittest.TestCase):
  def __init__():
    if not hasattr(unittest.TestCase, 'assertRaisesRegex'):
      setattr(unittest.TestCase, 'assertRaisesRegex', unittest.TestCase.assertRaisesRegexp)
```

Mingun:
Ok, I found a way to run python tests on Windows (7):
- Python 3.10 was already installed
- run `pip install xmlrunner` (because it was missed)
- change path delimiter from `:` to `;` in `ci-python`
- run `./ci-python` in Git Bash

I implemented the check in a common file which was placed to the `spec/python/extra` folder where other such files are located (should we rename `spec/python/extra` to `helpers/python`?).

Running tests reports two classes of warnings, I've fix them too.

PR #116, opened 2024-04-08
Rust: basic fixes of KST generator and update of tests

Mingun:
This PR is a small step to get working Rust target. It fixes tests generation that's now completely broken.

First of all, test files are generated in the `spec/rust` directory, but that directory is a Rust crate. Source code of crates (and unit tests) is located in the `src` subdirectory, sourse code of integration tests is located in `tests` directory. Because generated files more like a integration tests, now their generated is the `spec/rust/tests` directory.

The generated files had the following structure before this change:
```rust
// Autogenerated from KST: please remove this line if doing any edits by hand!

extern crate kaitai_struct;
extern crate rust;

use kaitai_struct::KaitaiStruct;
use rust::BcdUserTypeBe;

#[test]
fn test_bcd_user_type_be() {
    if let Ok(r) = BcdUserTypeBe::from_file("src/bcd_user_type_be.bin") {
        assert_eq!(r.ltr.as_int, 12345678);
        assert_eq!(r.ltr.as_str, "12345678");
        assert_eq!(r.rtl.as_int, 87654321);
        assert_eq!(r.rtl.as_str, "87654321");
        assert_eq!(r.leading_zero_ltr.as_int, 123456);
        assert_eq!(r.leading_zero_ltr.as_str, "00123456");
    }
}
```
It is incorrect in several places:
- first of all, if test file would not be found on expected place, the test actually test nothing and will pass. That is not we want. `if let Ok(r) = <Type>::from_file(...)` replaced by `let r = <Type>::from_file(...).expect(...)` to explicitly fail if binary file is not found
- then, the path to the binary files is incorrect. `cargo` run tests [in the crate's root](https://doc.rust-lang.org/cargo/commands/cargo-test.html#working-directory-of-tests) as working directory, which is `spec/rust/`. The relative path from this directory to directory with binary data is `../../src/`
- the path to the imported type is incorrect. The types is imported from the crate which is created from all generated `.rs` files + generated `lib.rs` (main library file) with content:
  ```rust
  pub mod <spec.id>;
  ```
  which means that all types will be available under name `<crate name>::<spec.id>::<TypeName>`. `<crate name>` is the `rust` (defined in `spec/rust/Cargo.toml`)
- crate renamed from very broad `rust` name to `kaitai_test_suite` and restricted from publishing
- crate updated to Rust 2021 edition to use modern Rust

In the end the generated test file would look like:
```rust
// Autogenerated from KST: please remove this line if doing any edits by hand!

use kaitai_struct::KaitaiStruct;
use kaitai_test_suite::bcd_user_type_be::BcdUserTypeBe;

#[test]
fn test_bcd_user_type_be() {
    let r = BcdUserTypeBe::from_file("../../src/bcd_user_type_be.bin").expect("file for parsing is not found");

    assert_eq!(r.ltr.as_int, 12345678);
    assert_eq!(r.ltr.as_str, "12345678");
    assert_eq!(r.rtl.as_int, 87654321);
    assert_eq!(r.rtl.as_str, "87654321");
    assert_eq!(r.leading_zero_ltr.as_int, 123456);
    assert_eq!(r.leading_zero_ltr.as_str, "00123456");
}
```


generalmimon:
@Mingun There is no point in you doing anything on Rust. I'm going to do it in the near future and will follow up on existing pull requests. You're just doing everything from scratch and wasting your time.

PR #117, opened 2024-04-10
KST: allow to generate `ks-debug` tests

Mingun:
This PR adds new property `debug` (default `false`) to KST tests, which allow to enable generation with `ks-debug` mode. The external call of `_read()` in tests is automatically generated
- make `DebugSwitchUser` test autogenerated
- implement test `DebugArrayUser` for Construct, Lua, Nim, Perl, and Rust


translator/src/main/scala/io/kaitai/struct/testtranslator/TestTranslator.scala
```diff
@@ -82,7 +82,7 @@ class TestTranslator(options: CLIOptions) {
         endian = None,
         bitEndian = None,
         encoding = None,
-        forceDebug = false,
+        forceDebug = debug,
```
Mingun, line 85: I wanted to name the new property as `auto-read` (with default to `true`), but because test is created using this code, more honestly to name it `debug`

PR #118, opened 2024-04-11
KST: Fix too small input file for `expr_io_ternary` test

Mingun:
The test requires `1 + 4 + 8 = 13` bytes, but `term_strz.bin` used as input contains only 12 bytes. Strange, but Ruby test was successfully passed -- it seems Ruby runtime has a bug in `substream` implementation. According to that test:
https://github.com/kaitai-io/kaitai_struct_ruby_runtime/blob/ecf67519a990296bb1e095363ef2dc414e600502/spec/subio_spec.rb#L70-L76
is it normal for Ruby to be able to read from a stream more that it contains successfully?

This commit replaces too small `term_strz.bin` file with `if_struct.bin` which size is 13 bytes.

The test was introduced in https://github.com/kaitai-io/kaitai_struct_tests/commits/dceaee2ebf9eb05177e6f5d0effd1b1b907be9ac

cc @GreyCat


generalmimon:
> Strange, but Ruby test was successfully passed -- it seems Ruby runtime has a bug in `substream` implementation.

Looks like it, that's why I added the `{Eof,Eos}ExceptionSized` tests in https://github.com/kaitai-io/kaitai_struct_tests/commit/99f9229714a29dc8353efecf5749688c0397b4ef.

generalmimon:
@Mingun:

> According to that test:
> https://github.com/kaitai-io/kaitai_struct_ruby_runtime/blob/ecf67519a990296bb1e095363ef2dc414e600502/spec/subio_spec.rb#L70-L76
> is it normal for Ruby to be able to read from a stream more that it contains successfully?

Well, it is, just as it is for `StringIO` or any other `IO` from the Ruby stdlib. If you look closely at the test you're linking to, it asserts that `Kaitai::Struct::SubIO` behaves the same as the Ruby built-in class `StringIO` in various cases.

Note that `Kaitai::Struct::SubIO` is not a class that any code outside the runtime library should directly interact with. It is only used internally in the `substream` method [`lib/kaitai/struct/struct.rb:529-533`](https://github.com/kaitai-io/kaitai_struct_ruby_runtime/blob/ecf67519a990296bb1e095363ef2dc414e600502/lib/kaitai/struct/struct.rb#L529-L533):

```rb
  def substream(n)
    sub = Stream.new(SubIO.new(@_io, @_io.pos, n))
    @_io.seek(@_io.pos + n)
    sub
  end
```

The problem is not with the internal `Kaitai::Struct::SubIO#read` method (as long it works the same as the native `read` methods would, it's fine), but that the `substream` method or the `Kaitai::Struct::SubIO` constructor lack any checks that we're not creating a substream that exceeds the parent stream size.

Mingun:
@GreyCat , @generalmimon, what the status?

generalmimon (approved):
@Mingun LGTM, thanks!

PR #119, opened 2024-04-11
Revert "Translator: allow go, but not by default" - Go translator is no worse than others

Mingun:
The Go translator is not included into `all` category since https://github.com/kaitai-io/kaitai_struct_tests/commit/0590480901f6074350800bd3b87b7420182b6cd7, which is strange. It's implementation not worse that, for example, Construct translator and it actually have more high rating on https://ci.kaitai.io.

### DIFF
--- a/translator/src/main/scala/io/kaitai/struct/testtranslator/Main.scala
+++ b/translator/src/main/scala/io/kaitai/struct/testtranslator/Main.scala
@@ -30,8 +30,6 @@ object Main extends App {
     "rust"
   )
 
-  val ALL_GOOD_LANGS = ALL_LANGS.filterNot(_ == "go")
-
   val INIT_OBJ_TYPE = "unique_top_level_container"
   val INIT_OBJ_NAME = "q1w2e3"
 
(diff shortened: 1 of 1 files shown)


generalmimon (approved):
Thanks! I've been a little annoyed for a while now that I had to type `-t all -t go` every time, but not enough to do anything about it.

@GreyCat mentioned that Go is not included in `-t all` because the KST translator used to crash due to "NotImplemented errors", see https://github.com/kaitai-io/kaitai_struct_tests/pull/75#issuecomment-584408559.

However, if I try `./spec_kst_to_all -t all --all-specs -f` now (with this PR's changes), it reaches all specs up to the last `zlib_with_header_78` without problems, even though there are 21 `scala.NotImplementedError`s when generating `.go` specs (20 times due to unimplemented [`doCast`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/27d502025c8b7b2ffeb9b5308c47c62fbda84047/shared/src/main/scala/io/kaitai/struct/translators/GoTranslator.scala#L277), once due to unimplemented [`testException`](https://github.com/kaitai-io/kaitai_struct_tests/blob/44ffc608052360b87d811f1804878188053ae616/translator/src/main/scala/io/kaitai/struct/testtranslator/specgenerators/BaseGenerator.scala#L30)). So I agree there is no reason to discriminate against Go anymore.

Mingun:
`testException` is implemented in #123

PR #121, opened 2024-04-11
Lua, Python: use special translation method for expected value of float checks as well

Mingun:
If it is used in `simpleEquality`, then it should be used in `floatEquality` as well


GreyCat:
Yep, makes sense. Thanks!

PR #122, opened 2024-04-11
Construct: support `exception` key top-level and in assertions

Mingun:
As a result, a bunch of tests was generated that checks exceptions. I checked in the code, that Construct does not wrap exceptions in its own exception classes. I didn't check that new tests work


generalmimon:
> I didn't check that new tests work

Well, that might be a problem.

I can tell you that this PR is useless in the current state, because the files generated by KSC for the `construct` target are meant to use only a completely separate [Construct](https://construct.readthedocs.io/en/latest/index.html) library, not our [Python runtime library](https://github.com/kaitai-io/kaitai_struct_python_runtime). Therefore, expecting Construct methods to raise exceptions like our Python target is nonsense, instead it should most likely be [something from here](https://construct.readthedocs.io/en/latest/api/exceptions.html).

Mingun:
It would be easier to test if CI could run these tests and show which exceptions would actually thrown. Feel free to edit this PR and use actual exceptions.

PR #124, opened 2024-04-17
formats_err: add tests for issue #534

Mingun:
This PR adds tests for https://github.com/kaitai-io/kaitai_struct/issues/534

Unfortunately, right now it contains Windows-specific paths which I'm not sure, should be completely removed or not? That is thy this PR in draft now

Related links:
- https://github.com/kaitai-io/kaitai_struct_compiler/pull/303 -- fix in the compiler
- https://github.com/kaitai-io/kaitai_struct_tests/pull/124 -- this PR
- https://github.com/kaitai-io/kaitai_struct/issues/534 -- issue

### DIFF
--- a/formats_err/imports_enum_leaking.ksy
+++ b/formats_err/imports_enum_leaking.ksy
@@ -1,0 +1,25 @@
+# imports/enum_two.ksy: /seq/0/enum:
+# 	error: unable to find type 'enum_one', searching from 'enum_two'
+#
+# imports/enum_two.ksy: /instances/instance_one/enum:
+# 	error: unable to find type 'enum_one', searching from 'enum_two'
+#
+meta:
+  id: imports_enum_leaking
+  imports:
+    - imports/enum_one
+    - imports/enum_two
+seq:
+  - id: another_one
+    type: u1
+    enum: enum_one::one
+  - id: another_two
+    type: u1
+    enum: enum_two::two
+instances:
+  instance_one:
+    value: 0
+    enum: enum_one::one
+  instance_two:
+    value: 0
+    enum: enum_two::two
--- a/formats_err/enum_unknown_inst_value.ksy
+++ b/formats_err/enum_unknown_inst_value.ksy
@@ -1,5 +1,5 @@
 # enum_unknown_inst_value.ksy: /instances/should_fail/enum:
-# 	error: unable to find enum 'animal', searching from enum_unknown_inst_value
+# 	error: unable to find enum 'animal', searching from 'enum_unknown_inst_value'
 #
 meta:
   id: enum_unknown_inst_value
--- a/formats_err/enum_unknown_seq.ksy
+++ b/formats_err/enum_unknown_seq.ksy
@@ -1,5 +1,5 @@
 # enum_unknown_seq.ksy: /seq/0/enum:
-# 	error: unable to find enum 'animal', searching from enum_unknown_seq
+# 	error: unable to find enum 'animal', searching from 'enum_unknown_seq'
 #
 meta:
   id: enum_unknown_seq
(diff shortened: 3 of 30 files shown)


formats_err/imports_enum_leaking.ksy
```diff
@@ -0,0 +1,15 @@
+# ..\tests\formats_err/imports/enum_two.ksy: /seq/0/enum:
```
Mingun, line 1: Should this be
```suggestion
# imports/enum_two.ksy: /seq/0/enum:
```
?
Or
```suggestion
# ../tests/formats_err/imports/enum_two.ksy: /seq/0/enum:
```
?

generalmimon: The first option, because that's what you get on Unix-like systems. The fact that it's not the case on Windows can be considered a minor bug (although it's just a detail in the error message, so it's a bit annoying, but relatively harmless).

I also develop mostly on Windows (I run some things in WSL 2, but not all), so this sort of affects my local setup as well, but I'm relatively fine with ignoring a few local path-related test failures. If you want to resolve this properly, check out this commit of mine first, which will give you an explanation how these paths in `formats_err` error messages work: https://github.com/kaitai-io/kaitai_struct_compiler/commit/5bc871a79afc693ca08bf6109416ce019a919cb1

The thing is that the Unix path would only contain forward slashes (like `../tests/formats_err/imports/enum_two.ksy: /seq/0/enum:`), in which case `../tests/formats_err/` gets removed and only `imports/enum_two.ksy: /seq/0/enum:` remains. After taking a quick look, this apparently doesn't happen on Windows because of the way how `yamlDir` is computed here in KSC ([`JavaKSYParser.scala:21-22`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/2191070c7a8a7e0c19b01ca79f5616437cc7bc6f/jvm/src/main/scala/io/kaitai/struct/formats/JavaKSYParser.scala#L21-L22)):

```scala
      val yamlDir = Option(new File(yamlFilename).getParent).getOrElse(".")
      val specs = new JavaClassSpecs(yamlDir, config.importPaths, firstSpec)
```

The Java `File` class notoriously _normalizes_ all paths to backslash-separated `\` on Windows (more on that here: https://github.com/kaitai-io/kaitai_struct/issues/507#issuecomment-1679686055), so it isn't very surprising to me that the value of `yamlDir` is `..\tests\formats_err` even though `yamlFilename` was `../tests/formats_err/imports_enum_leaking.ksy`.

The fix would be to use some other way to get the directory name of `yamlFilename` than `new File(yamlFilename).getParent` (as far as I know, the path normalization of Java `File` class can't be disabled, but there may be other APIs that don't do it). I'll just note that we shouldn't use any custom string manipulation for this (because we'd probably mess it up), but really try to find some existing reliable API in the standard library first.

Mingun: Ok, I removed the prefix and fix stripping on Windows in https://github.com/kaitai-io/kaitai_struct_compiler/pull/304

generalmimon (changes requested):
Please make all these lines at https://github.com/kaitai-io/kaitai_struct_tests/pull/124.patch disappear:

```
\ No newline at end of file
```

Mingun:
@generalmimon ?

formats_err/imports_enum_leaking.ksy
```diff
@@ -0,0 +1,25 @@
+# imports/enum_two.ksy: /seq/0/enum:
+# 	error: unable to find type 'enum_one', searching from 'enum_two'
+#
+# imports/enum_two.ksy: /instances/instance_one/enum:
+# 	error: unable to find type 'enum_one', searching from 'enum_two'
+#
```
Mingun, line 6: Strange, but the second error never reported in this test, even when I apply https://github.com/kaitai-io/kaitai_struct_compiler/pull/313, so all errors should be reported. Don't figure yet, why

Mingun (commented):
Rebased on top of #130 (which includes #129) because fix in compiler was rebased on top of corresponding compiler PRs

PR #126, opened 2024-04-29
Julia

rystidia:
Add tests for compiled tests, running scripts, and KST translator support for Julia

This PR addresses the issue https://github.com/kaitai-io/kaitai_struct/issues/1109


rystidia:
I created [this PR](https://github.com/kaitai-io/kaitai_struct_docker_images/pull/1) (which was merged). It worked for me when running like `./docker-ci -t julia -i 1.9`. Should I somehow change it?

GreyCat:
Nope, sorry, please disregard my previous comment — I've just overlooked that you've already created it. Now it's merged and this enables to test tests locally.

translator/src/main/scala/io/kaitai/struct/testtranslator/specgenerators/JuliaSG.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  override def noAsserts() =
+    out.puts("")
+
+  def translateAct(x: Ast.expr) =
+    translator.translate(x).replace("this." + Main.INIT_OBJ_NAME, "r")
+
+  def translateExp(x: Ast.expr) =
+    translator.translate(x).replace("this._root", className)
```
generalmimon, line 74: I don't like seeing this pattern at all since:

1. it uses a textual replacement... (which is inherently wrong and should be avoided as much as possible - textual replacement can be only ever appropriate in free-form text, not in anything that has a certain format that must be followed, like the code expression of some programming language. The existing `.replace("this." + Main.INIT_OBJ_NAME, "r")` is also fundamentally flawed, but at least it is very unlikely to cause any actual problems, because `val INIT_OBJ_NAME = "q1w2e3"` is sufficiently "random" that it probably won't clash with anything unintended.)
2. with something that doesn't even make sense at first sight (in general you can hardly "replace" `_root`, which is the reference to the root **instance** of the top-level class, with the **class** name itself. That is like apples and oranges).

This shouldn't be necessary, and I'm afraid it indicates some issues with the compiler code generation...

rystidia:
@generalmimon Thanks for your detailed review. As a reaction to your suggested changes, I opened [!127](https://github.com/kaitai-io/kaitai_struct_tests/pull/127)

PR #140, opened 2026-04-18
Adapt and extend the test suite to cover the new `enums` syntax

generalmimon:
See https://github.com/kaitai-io/kaitai_struct/issues/1288

This accompanies the changes in https://github.com/kaitai-io/kaitai_struct_compiler/pull/335.


formats/docstrings.ksy
```diff
@@ -21,6 +21,12 @@ types:
       And the text continues after that. Here's a MarkDown-style link:
       [woohoo](http://example.com) - one day it will be supported as
       well.
+enums:
+  sample_enum:
+    doc: A detailed description for enum "sample_enum"
+    type: u1
+    values:
+      1: foo
```
Mingun, line 29: Add docstring for value too?

generalmimon: @Mingun:

> Add docstring for value too?

I thought about it, but since that's already covered in `formats/enum_fancy.ksy`:

https://github.com/kaitai-io/kaitai_struct_tests/blob/fbed81935dec1b4b49f7a2a3abb409f69ec31af8/formats/enum_fancy.ksy#L11-L16

..., I don't think it's necessary to do it here as well.

Mingun: Then why not just add an enum comment to `enum_fancy`? Now it looks like dividing tests according to an incomprehensible area of responsibility - in docstrings.ksy we seem to be testing all documenting keys, but not in enum value; in enum_fancy.ksy we test everything related to enum, but not enum documentation...

generalmimon:
The comment https://github.com/kaitai-io/kaitai_struct_compiler/pull/335#issuecomment-4361793075 also applies here - this is finished, but we don't yet have a `2.x` branch to merge it into, see https://github.com/kaitai-io/kaitai_struct/issues/1293.

PR #144, opened 2026-06-07
C++: Use the same precision 1e-6 as for other languages for floating-point checks

Mingun:
It seems that current 1e-4 was chosen arbitrately in 224ae5c90c68084d3fd7f920663ea105dd31f77a. I do not see any needs to use more tolerant check for C++ in comparison to other languages. For example, Lua and JavaScript uses `double` for representing numbers and they making checks using tolerance of 1e-6.

### DIFF
--- a/spec/cpp_stl_11/test_expr_array.cpp
+++ b/spec/cpp_stl_11/test_expr_array.cpp
@@ -17,10 +17,10 @@ BOOST_AUTO_TEST_CASE(test_expr_array) {
     BOOST_CHECK_EQUAL(r->aint_min(), 49185);
     BOOST_CHECK_EQUAL(r->aint_max(), 1123362332);
     BOOST_CHECK_EQUAL(r->afloat_size(), 3);
-    BOOST_CHECK_CLOSE(r->afloat_first(), -2.6839530254859364E-121, 1e-4);
-    BOOST_CHECK_CLOSE(r->afloat_last(), -1.1103359815095273E-175, 1e-4);
-    BOOST_CHECK_CLOSE(r->afloat_min(), -8.754689149998834E+288, 1e-4);
-    BOOST_CHECK_CLOSE(r->afloat_max(), -1.1103359815095273E-175, 1e-4);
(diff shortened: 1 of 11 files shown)


generalmimon:
@Mingun:

> It seems that current 1e-4 was chosen arbitrately in 224ae5c90c68084d3fd7f920663ea105dd31f77a.

The reason I chose `1e-4` was that Boost.Test accepts the tolerance in _percentage units_ (see https://www.boost.org/doc/libs/1_81_0/libs/test/doc/html/boost_test/utf_reference/testing_tool_ref/assertion_boost_level_close.html):

> [`BOOST_<level>_CLOSE`](https://www.boost.org/doc/libs/1_81_0/libs/test/doc/html/boost_test/utf_reference/testing_tool_ref/assertion_boost_level_close.html)
> ```
> BOOST_WARN_CLOSE(left, right, tolerance);
> BOOST_CHECK_CLOSE(left, right, tolerance);
> BOOST_REQUIRE_CLOSE(left, right, tolerance);
> ```
> 
> (...)
> 
> (...) Last third parameter defines the tolerance for the comparison in [percentage units](https://www.boost.org/doc/libs/1_81_0/libs/test/doc/html/boost_test/testing_tools/extended_comparison/floating_point.html).

So my thinking was that in order to get the effective tolerance of `1e-6` as a fraction (i.e. not in percentage units), I need to pass `1e-4` to Boost, because Boost will divide it by 100 (and `1e-4 / 100 == 1e-6`).

It seems that we could also use [`BOOST_<level>_CLOSE_FRACTION`](https://www.boost.org/doc/libs/1_81_0/libs/test/doc/html/boost_test/utf_reference/testing_tool_ref/assertion_boost_level_close_fraction.html) instead, which I wasn't aware of at the time of making that commit. I assume `BOOST_<level>_CLOSE_FRACTION(left, right, 1e-6)` is equivalent to `BOOST_<level>_CLOSE(left, right, 1e-4)`.

---

However, the problem here is deeper than value of `1e-4` vs. `1e-6` - in Boost.Test, "tolerance" actually means something different than in rest of the languages. I opened an issue for this: https://github.com/kaitai-io/kaitai_struct/issues/1314

